PR #2, opened 2023-09-11
Add Dockerfile

JosseVanDelm:
Courtesy of @nazavode

This dockerfile does the following:

Stage 1:  Inside the snitch_cluster container:
1) Download the latest  `https://github.com/kuleuven-micas/snitch_cluster.git` and build it with verilator.
2) compile `spike-dasm` and install
3) compile `snitch runtime`
4) download snitch-specifc versions of clang/llvm/lld
5) download and compile python 3.11

Stage 2: Inside the `ubuntu 18.04` container:
1) copy-paste all binaries built in the previous stage
2) install python's ubuntu dependencies
3) install git
4) install a pre-built mlir (version 16.0.6), available as `mlir-opt-16` inside the container, optimized build, without assertions
5) install make

I'm currently/temporarily providing a public build of this dockerfile here for you to try out:
```
docker pull ghcr.io/jossevandelm/snax-toolchain
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,19 @@ Repository for running MLIR compiled stuff on SNAX
 ## Requirements
 
 [stardew](https://github.com/Groverkss/stardew)
+
+## Docker Container
+
+For compiling the low-level kernels you need the snitch compilation toolchain, 
+which is easiest to get through a docker container.
+A Dockerfile for such container is provided here:
+```sh
+cd container
+docker build . -t snax-toolchain
+cd ..
+```
+Then you can run your experiments in this repository with:
+```sh
+docker run -itv `pwd`:/repo:z snax-toolchain
+```
+The repository will be available under `/repo`
--- a/README.md
+++ b/README.md
@@ -10,10 +10,10 @@ Repository for running MLIR compiled stuff on SNAX
 
 For compiling the low-level kernels you need the snitch compilation toolchain, 
 which is easiest to get through a docker container.
-A Dockerfile for such container is provided here:
+A Dockerfile for such container is provided here (note that leaving out the `config` `--build-arg` will use the default `snitch_cluster` setup:
 ```sh
 cd container
-docker build . -t snax-toolchain
+docker build . -t snax-toolchain --build-arg config=path_to_your_hjson_file.hjson
 cd ..
 ```
 Then you can run your experiments in this repository with:
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -4,9 +4,12 @@ ARG config=target/snitch_cluster/cfg/default.hjson
 
 FROM ghcr.io/pulp-platform/snitch_cluster:latest@sha256:aa0d9a25022d1ef51f451dc667b850e3e8dba4c15d949b51b5cfde10f89defe9 as builder
 
-RUN git clone --recursive https://github.com/kuleuven-micas/snitch_cluster.git /src \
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -5,3 +5,19 @@ Repository for running MLIR compiled stuff on SNAX
 ## Requirements
 
 [stardew](https://github.com/Groverkss/stardew)
```
rgantonio, line 7: Just to clarify, when you meant this is the requirement, you meant we also have to follow the instructions from stardew right?

container/Dockerfile
```diff
@@ -0,0 +1,87 @@
+# Courtesy of Federico Ficarelli
+
+FROM ghcr.io/pulp-platform/snitch_cluster:latest@sha256:9978f9d0e3e6b7ad1b2c1f0eb628a113226e8443bc42ae8b4ddd3f34d9528e96 as builder
+
+RUN git clone --recursive https://github.com/kuleuven-micas/snitch_cluster.git /src \
+ # verilator model
+ && cd /src/target/snitch_cluster \
+ && make bin/snitch_cluster.vlt
```
rgantonio, line 8: Just a thought, but what if we have different builds? I mean, what if we can reconfigure the .hjson file to different architectures. With SNAX, no SNAX, 1 SNAX 1 Snitch and 1 DMA and so on. I wonder if it's possible if we can have a switcher for that. Right now, it's not really needed but I foresee we will need that in the future.

rgantonio (commented):
I guess this is fine :) I just have some questions for clarification.

rgantonio:
An idea came to mind, I am currently testing the build instructions from scratch. I wonder if it is instead better to just provide the built binary? I checked it was like 32 MB. Is it normal if people would just upload such builds to git? Or is that a bad idea?

It may still be smaller than manually building containers or pre-built containers? What do you think?

README.md
JosseVanDelm: This is from a previous PR, but yes @jorendumoulin is going to add some more instructions on that later

container/Dockerfile
JosseVanDelm: Good call, I can add a docker build arg there.
https://vsupalov.com/docker-env-vars/

JosseVanDelm:
@rgantonio we should NEVER put binary files in git, since git does not recognize them and they will be kept in the git history forever. Having many binary files in the history will make downloading the repo much slower and will make git much slower to work with, so we should avoid that at all times.
If we want to keep these binaries somewhere in the CI we should either use caching or artefacts, but not  in the repo itself.

container/Dockerfile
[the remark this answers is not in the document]
JosseVanDelm: Thanks for reporting, I'll investigate

JosseVanDelm: Added in [3d830ef](https://github.com/KULeuven-MICAS/snax-mlir/pull/2/commits/3d830eff8e1d08d60b34b98db4108399156ed987)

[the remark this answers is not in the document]
JosseVanDelm: The reason why this failed was because we updated the main branch of the repository, and the Dockerfile still linked to an old starter container, new versions are fixed now

JosseVanDelm:
I've added a few things:
* build argument to make sure you can use multiple configurations
* a CI check to make sure it builds, if it is changed.
* the version of the snitch_cluster repo is now fixed

jorendumoulin (approved):

PR #6, opened 2023-09-14
Add code formatting ci & pre-commit hooks

jorendumoulin:
This commit adds pre-commit code formatting hooks, just like in [snax-dev](https://github.com/KULeuven-MICAS/snax-dev/pull/3)


requirements.txt
```diff
@@ -0,0 +1,3 @@
+pre-commit
+ruff
+black
```
JosseVanDelm, line 3: Wondering whether we should include `clang-format` here too, or should remove black and ruff here and just support them as pre-commit hooks (I believe pre-commit installs them by itself, no?)

JosseVanDelm (approved):
Just on little nit. Other than that this is :fire:

requirements.txt
jorendumoulin: To avoid duplicate requirements in .pre-commit-config.yaml and requirements.txt, I feel like its best to just remove them from requirements.txt

PR #8, opened 2023-09-15
docs: Add tracing, disassembly instructions, harmonize docker image naming

JosseVanDelm:
This adds some instructions to perform dissasembly, tracing and should make it easier for newcomers to get started.

In terms of tutorial this is not yet a full blown tutorial ( I think it is too early for that ).
A proper tutorial would have examples etc.


jorendumoulin (approved):
Nice documentation!

PR #9, opened 2023-10-31
Add XDSL Compiler Flow

jorendumoulin:
This pull request adds the XDSL project as a submodule to write custom MLIR passes for lowering linalg->snax.
It also includes a first simple pass that detects elementwise integer multiplications, and lowers these to an external function call. This external function call can then be an implementation using a snax accelerator.


compiler/transforms/allocate_elementwise_mult.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: linalg_ext.Mul, rewriter: PatternRewriter):
+        if op.library_call is None:
+            return
+
+        rewriter.replace_matched_op(func.Call(op.library_call.data, op.operands, []))
+
+        return
+
```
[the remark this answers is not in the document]
jorendumoulin: Why is this wrong? This pattern matches on every linalg.mul operation with a library call, and replaces the linalg.mul op with a func.call to the specified library call? A subsequent pass then inserts the external functions

compiler/transforms/allocate_elementwise_mult.py
```diff
@@ -0,0 +1,101 @@
...
+
+class AddExternalFunc(RewritePattern):
```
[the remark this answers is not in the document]
JosseVanDelm: Yes, Anton recently implemented a mechanism that  doesn't insert an external function call declarartion if one is already there, which seems useful here :)

compiler/dialects/linalg_ext.py
```diff
@@ -0,0 +1,72 @@
...
+                    sizes.append(dim)
+        for output in self.outputs:
+            if isinstance(output.type, ShapedType):
+                for dim in output.type.get_shape():
+                    sizes.append(dim)
+        return sizes
+
+
+# Extended Linalg Dialect
+LinalgExt = Dialect("linalg_ext", [Mul])
```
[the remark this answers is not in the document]
jorendumoulin: Changed this to linalg_extension. The file extends the linalg dialect in xDSL to support more operations. To me this seems like a better option until we maybe can upstream the linalg extensions upstream.

compiler/transforms/allocate_elementwise_mult.py
```diff
@@ -0,0 +1,101 @@
...
+    Looks for hwpe function calls and adds an external
+    func call to it for LLVM to link in
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, module: builtin.ModuleOp, rewriter: PatternRewriter):
+        for op in module.walk():
+            if not isinstance(op, func.Call):
+                continue
+            if "snax_hwpe" not in op.callee.string_value():
```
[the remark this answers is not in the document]
jorendumoulin: I agree, but for now i would leave it like this.
Removing the check will insert a external function definition for every function call (not just the ones we just inserted), which may lead incorrect behaviour.

 
A following PR which does the following:
- Split pass in two:
    * A pass which adds a library call to linalgs compatible with a certain accelerator
    * A pass which replaces all linalgs with library calls with an external function call
    
would solve this as the second pass is more general and would handle these special cases

compiler/transforms/allocate_elementwise_mult.py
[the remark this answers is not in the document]
JosseVanDelm: Oh no of course you want to replace the function if the library call is not empty, makes perfect sense!

compiler/interpreters/linalg_extension.py
JosseVanDelm, line 1: Are we using/testing this anywhere?
Otherwise I would delete this, at least for now

tests/filecheck/lit.cfg
```diff
@@ -0,0 +1,12 @@
...
+
+config.test_source_root = os.path.dirname(__file__)
+snax_src = os.path.dirname(os.path.dirname(config.test_source_root))
+
+config.name = "SNAX"
+config.test_format = lit.formats.ShTest(preamble_commands=[f"cd {snax_src}"])
+config.suffixes = ['.test', '.mlir', '.py']
+
+config.substitutions.append(('XDSL_ROUNDTRIP', "snax-opt %s --print-op-generic --split-input-file | snax-opt --split-input-file | filecheck %s"))
+config.substitutions.append(("XDSL_GENERIC_ROUNDTRIP", "snax-opt %s --print-op-generic --split-input-file | filecheck %s --check-prefix=CHECK-GENERIC"))
```
JosseVanDelm, line 12: You are not using these in your tests, I think it's useful to have them here, but they should be tested/used

tests/filecheck/transforms/dispatch_elementwise_mult.mlir
```diff
@@ -0,0 +1,23 @@
+// RUN: ./compiler/snax-opt %s -p dispatch-elementwise-mult --allow-unregistered-dialect --print-op-generic | filecheck %s
```
JosseVanDelm, line 1: Is this how that works? I'm not sure if we need the GENERIC roundtrip here instead maybe.
```suggestion
// RUN: XDSL_ROUNDTRIP
```

compiler/snax-opt
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python3
+
+import re
+import sys
+from tools.snax_opt_main import main
+
+if __name__ == "__main__":
+    sys.argv[0] = re.sub(r"(-script\.pyw|\.exe)?$", "", sys.argv[0])
+    sys.exit(main())
```
JosseVanDelm, line 9: Is this necessary?

JosseVanDelm (changes requested):

tests/filecheck/transforms/dispatch_elementwise_mult.mlir
jorendumoulin: also added a roundtrip test

compiler/interpreters/linalg_extension.py
jorendumoulin: deleted

JosseVanDelm (approved):
:green_heart:

PR #11, opened 2023-11-08
xDSL: Split Lowering Passes

jorendumoulin:
This pull request does a couple of things:

- It splits the dispatch-elementwise-mult pass into two passes:
    - dispatch-elementwise-mult: this pass detects the elementwise multiplications compatible with the hwpe accelerator, and adds a library call on succesfull hits
    - linalg-to-library-call: this pass simply matches on every linalg which has a library call, and inserts the external function and a function call
- It reverts the linalg extension to be able to use linalg named ops, so we only use linalg generic from now.
    - If we were to use linalg named ops, we would need to implement the conversion of named ops -> generic ops to be able to retrieve the indexing maps of the operation
    - Using linalg.generic, the problem of detecting certain operations becomes somewhat more difficult, but I think this is manegable because:
        - ZigZag does also not contain named ops the constraining mechanism to be used there will be based on the type of indexing maps and operation. The indexing maps are directly available from linalg generic ops. The operation must be detected from the block of the linalg.generic operation, but for accelerators this is limited to (mult, add, multiply-accumulate)


compiler/transforms/dispatch_elementwise_mult.py
```diff
@@ -1,26 +1,29 @@
...
     RewritePattern,
     op_type_rewrite_pattern,
 )
-from xdsl.traits import SymbolTable
 
 
 class AddLibraryCall(RewritePattern):
     @op_type_rewrite_pattern
-    def match_and_rewrite(self, op: linalg_extension.Mul, rewriter: PatternRewriter):
+    def match_and_rewrite(self, op: linalg.Generic, rewriter: PatternRewriter):
```
JosseVanDelm, line 15: Would be good to have some part of your reasoning in your PR here directly?
```suggestion
    def match_and_rewrite(self, op: linalg.Generic, rewriter: PatternRewriter):
        """Add a library call to a linalg generic that implements an elementwise multiplication (similar to linalg.mul).    Adding such an attribute to a linalg.mul op directly is not possible."""
```

tests/filecheck/transforms/linalg-to-library-call.mlir
JosseVanDelm, line 1: Maybe you can add one more linalg generic with the same attribute to this, to check if only one func.func gets generated?

compiler/transforms/dispatch_elementwise_mult.py
```diff
@@ -35,67 +38,46 @@ def match_and_rewrite(self, op: linalg_extension.Mul, rewriter: PatternRewriter)
...
+        mult_op = op.body.block.first_op
+        yield_op = op.body.block.last_op
 
-    @op_type_rewrite_pattern
-    def match_and_rewrite(self, op: linalg_extension.Mul, rewriter: PatternRewriter):
-        if op.library_call is None:
+        # last operation is linalg.yield
+        if not isinstance(yield_op, linalg.YieldOp):
+            return
+        # first operaion is arith.muli
```
JosseVanDelm, line 56:
```suggestion
        # first operation is arith.muli
```

compiler/transforms/dispatch_elementwise_mult.py
```diff
@@ -35,67 +38,46 @@ def match_and_rewrite(self, op: linalg_extension.Mul, rewriter: PatternRewriter)
...
-    """
+        mult_op = op.body.block.first_op
+        yield_op = op.body.block.last_op
 
-    @op_type_rewrite_pattern
-    def match_and_rewrite(self, op: linalg_extension.Mul, rewriter: PatternRewriter):
-        if op.library_call is None:
+        # last operation is linalg.yield
+        if not isinstance(yield_op, linalg.YieldOp):
+            return
```
JosseVanDelm, line 55: I don't think it's required to test this.
A linalg generic's body is required to have a yield operation, because otherwise it should trip the verifier (has no terminator)

JosseVanDelm: Oh I think I see why, you need this for your later test I guess?

PR #12, opened 2023-11-09
end-to-end workflow

jorendumoulin:
This pull request implements an end-to-end flow from mlir to simulation on a snitch cluster. The steps taken look like this:

Preprocess the input mlir to with generic printing and local scope
`input.mlir ->  input.preproc.mlir`

Translate MLIR 16 to MLIR18 with a python script
`input.preproc.mlir -> input.preproc18.mlir`

Apply the custom `snax-opt` compiler. This includes passes to lower the linalg generics to external function calls which drive the accelerator
`input.preproc18.mlir -> input.snax-opt18.mlir`

Translate back to MLIR 16
`input.snax-opt18.mlir -> input.snax-opt.mlir`

Now use the existing MLIR workflow to lower and translate the remaining MLIR to LLVM, compile and run

### DIFF
--- a/runtime/tomlir16.py
+++ b/runtime/tomlir16.py
@@ -1,0 +1,15 @@
+#!/usr/bin/env python3
+
+import re
+import sys
+
+
+if __name__ == "__main__":
+    ir = sys.stdin.read()
+    ir = re.sub(
+        r"<({.*?})>\s*(\({[^}]*}\))", lambda x: f"{x.group(2)} {x.group(1)}", ir
+    )
+    ir = re.sub('"indexing_maps"', "indexing_maps", ir)
+    ir = re.sub('"iterator_types"', "iterator_types", ir)
+    ir = re.sub('"operandSegmentSizes"', "operand_segment_sizes", ir)
+    print(ir)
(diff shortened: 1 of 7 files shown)


runtime/tomlir16.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3
+
+import re
+import sys
+
+
+if __name__ == "__main__":
+    ir = sys.stdin.read()
+    ir = re.sub(
```
JosseVanDelm, line 9: A comment to explain what happens here would be nice for future reference!

runtime/tomlir16.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3
+
+import re
+import sys
+
```
JosseVanDelm, line 5: Add a comment with list of all the things you need to change and why

runtime/tomlir18.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python3
+
+import re
+import sys
+
```
JosseVanDelm, line 5: Add a comment with list of all the things you need to change and why

JosseVanDelm (approved):
Looks very good!
We need some more comments though!

PR #16, opened 2023-11-15
compiler transform: set memory spaces

jorendumoulin:
This PR initializes the first pass to raise DMA calls for snax to MLIR: -set-memory-spaces

This pass makes sure that the memrefs used in the code are all assigned to a memory space.
For now this is based on simple rules:

- Memref arguments of a public function (which can be called from C code), that do not yet have an assigned memory space, are defaulted to memory space `0`.
- Linalgs can only operate on memrefs in memory space `1`.

The transformation happens in 3 steps:

1. `InitFuncMemorySpace`
In this step, all memrefs arguments of a function are assigned to a memory space. The entire function then uses this memref
2. `InitLinalgMemorySpace`
In this step, the arguments used by linalg.Generic operations are converted to be memrefs in memory space 1. Conversions are handled by `memref.memory_space_cast` operations.
3. `RealizeMemorySpaceCasts`
In this step, the memory space casts are realized by transforming them to `memref.alloc` and `memref.copy`. Some simple dependency analysis is present to determine when and if a certain memref must be copied from memory space `0` to `1` or the other way around.

### DIFF
--- a/compiler/transforms/set_memory_space.py
+++ b/compiler/transforms/set_memory_space.py
@@ -151,13 +151,6 @@ class RealizeMemorySpaceCasts(RewritePattern):
 
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: memref.MemorySpaceCast, rewriter: PatternRewriter):
-        # dont rewrite for invalid memory space casts:
-        # if these are still present, dce should remove them later
-        if len(op.results) < 1:
-            return
-        if len(op.results[0].uses) == 0:
-            return
-
         # check if result is memreftype to make pyright happy
         if not isinstance(op.results[0].type, memref.MemRefType):
             return
--- a/compiler/transforms/set_memory_space.py
+++ b/compiler/transforms/set_memory_space.py
@@ -1,7 +1,9 @@
-from xdsl.dialects import func, builtin, memref
+from xdsl.dialects import func, builtin, memref, linalg
 from xdsl.ir import MLContext
+from xdsl.ir.core import SSAValue
 from xdsl.passes import ModulePass
 from xdsl.pattern_rewriter import (
+    GreedyRewritePatternApplier,
     RewritePattern,
     PatternRewriter,
     op_type_rewrite_pattern,
@@ -12,7 +14,7 @@ from xdsl.pattern_rewriter import (
 class AddMemorySpace(RewritePattern):
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter):
-        """Add a default L3 memory space to memrefs used in the function
+        """Add a default (0 : i32) memory space to memrefs used in the function
         that do not have a memory space specified yet"""
 
         # Function must be public
@@ -35,7 +37,7 @@ class AddMemorySpace(RewritePattern):
         ):
             return
 
-        # Mapping functino to assign default memory space 0
+        # Mapping function to assign default memory space (0 : i32)
         def change_to_memory_space(t):
             if isinstance(t, memref.MemRefType):
                 if isinstance(t.memory_space, builtin.NoneAttr):
@@ -47,7 +49,8 @@ class AddMemorySpace(RewritePattern):
                     )
             return t
 
-        # Define new funcion type with updated inputs & outputs
+        # Define new function type with updated inputs and outputs
+        # mapped to a default memory space
         new_function_type = builtin.FunctionType.from_lists(
             map(change_to_memory_space, op.function_type.inputs),
             map(change_to_memory_space, op.function_type.outputs),
@@ -57,7 +60,7 @@ class AddMemorySpace(RewritePattern):
         for arg in op.args:
             arg.type = change_to_memory_space(arg.type)
 
-        # Define op with new function type and copy region contents
+        # Define new function op with new type and copy region contents
         new_op = func.FuncOp(
             op.sym_name.data,
             new_function_type,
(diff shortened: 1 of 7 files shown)


tests/filecheck/transforms/set-memory-space.mlir
```diff
@@ -0,0 +1,42 @@
+// RUN: ./compiler/snax-opt %s -p set-memory-space --allow-unregistered-dialect --print-op-generic | filecheck %s
```
JosseVanDelm, line 1: Why is --allow-unregistered-dialect necessary here, which operations are not registered??

compiler/transforms/set_memory_space.py
```diff
@@ -0,0 +1,226 @@
...
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: memref.MemorySpaceCast, rewriter: PatternRewriter):
+        # check for invalid memory space cast,
+        # no results => unused, removed by cse
+        if len(op.results) < 1:
+            return
+
+        # remove cast, unused op: will be removed by cse
+        if len(op.results[0].uses) == 0:
```
JosseVanDelm, line 160: I don't understand this part.
Why is this necessary?
Do you mean that unused parts will be removed? Because then you are running dce (dead code elimination) instead of common subexpression elimination (CSE), no?
I believe DCE is part of `--canonicalize`, but I don't see it used in any of your tests.

compiler/transforms/set_memory_space.py
```diff
@@ -0,0 +1,226 @@
...
+            new_function_type,
+            region=rewriter.move_region_contents_to_new_regions(op.regions[0]),
+            visibility=op.sym_visibility,
+        )
+
+        # Replace function op
+        rewriter.replace_matched_op(new_op)
+
+
+class InitLinalgMemorySpace(RewritePattern):
```
JosseVanDelm, line 68: Why do we need to insert a `memory_space_cast` (runtime cast) if we change all `memory_space_casts` in the following pass to be allocs anyway?
I feel like there is only use for this pass if afterwards some `memory_space_cast` will remain, which is not the case I believe?
Is there any `memory_space_cast` that eventually will be turned into risc-v code?

tests/filecheck/transforms/set-memory-space.mlir
```diff
@@ -0,0 +1,42 @@
...
+
+"builtin.module"() ({
```
JosseVanDelm, line 3: I feel like this test is quite complex, since it tests many things in sequence, do you think it would make sense to split up the tests?

compiler/transforms/set_memory_space.py
```diff
@@ -0,0 +1,226 @@
...
+                is_input = True
+            else:
+                is_input = op.results[0] in use_op.inputs
+            if is_input:
+                # insert copy op
+                copy_op = memref.CopyOp(op.source, op.dest)
+                rewriter.insert_op_before(copy_op, use_op)
+                break
+
+        # insert copy from for first use as output
```
JosseVanDelm, line 198:
```suggestion
        # insert "copy from" for first use as output
```

compiler/transforms/set_memory_space.py
```diff
@@ -0,0 +1,226 @@
...
+            op.results[0].type.get_shape(),
+            op.results[0].type.layout,
+            op.results[0].type.memory_space,
+        )
+
+        # Insert copy ops if newly allocated memref is used as
+        # input or output, list to visit all uses of allocated memrefs:
+        uses = [x.operation for x in op.results[0].uses]
+
+        # insert copy to for first use as input
```
JosseVanDelm, line 180:
```suggestion
        # insert "copy to" for first use as input
```
Reads a bit weird otherwise

JosseVanDelm (changes requested):
Cool! Please address my questions!

compiler/transforms/set_memory_space.py
jorendumoulin: okay, I thought DCE was part of CSE, but I could be mistaken.
These checks are mainly to make sure the pass does not apply to ops without results because the following lines would then give nasty errors:
```python
op.results[0].type.get_element_type(),
```
I'll update the comments to make this more clear

compiler/transforms/set_memory_space.py
jorendumoulin: Yes! Good question!

I want to stay flexible with memory casts, consider the following scenario:
There is a global address space from address 0 to FF. There are three memory spaces shown in the diagram: 5 - 6 - 8. I want them to be able to overlap each other (In snitch this case may exist for the DMA core which can access L1+L3 while an accelerator only has access to L3)

Now consider the data p1, present in memory space 8 (p1, 8)
If we want to cast it to memory space 6, this is not possible. Therefore, we must allocate + copy. However, in memory space 5 this is not necessary as this memory space already has access to p1. 

This second use case is not yet implemented, but the first conversion to memory_space_casts and only later realizing them allows for this flexibility

```
             ▲
             │xFF
       ┌──   │
       │     │
       │     │
       │8    │
┌──    │     │
│      │     ├──────  p1
│      └──   │
│            │
│5     ┌──   │
│      │     │
│      │     │
│      │     │
│      │     │
└──    │6    │
       │     │
       │     │
       │     │
       └──   │
             │
             │
             │
             │ x00
```

JosseVanDelm: I think we can technically still leave them out, but okay to leave them here!

compiler/transforms/set_memory_space.py
jorendumoulin: Turns out it is indeed impossible to have a `memref.memory_space_cast` without results, so this test case is not possible. The check is therefore also not necessary, and has been removed

JosseVanDelm (approved):
Great!

PR #17, opened 2023-11-15
bump xdsl

jorendumoulin:


JosseVanDelm (approved):
LGTM, maybe we should look into using the setup python action instead at some point?

PR #28, opened 2023-11-22
Add conversion for memref.alloc to _mlir_memref_to_llvm_alloc

JosseVanDelm:


kernels/alloc/main.c
jorendumoulin, line 1: This test does not actually check if the allocation is successful i think? It prints out the results, but does not check if they are equal. If the allocation\synchronisation fails, this goes unnoticed in CI. Maybe write some data with core 1 and then check the data with core 2 to determine if they have access to the same allocated buffer?

runtime/tomlir16.py
```diff
@@ -18,8 +18,12 @@
 
     # swap attributes and region from place, and remove <> from attributes
     ir = re.sub(
-        r"<({.*?})>\s*(\({[^}]*}\))", lambda x: f"{x.group(2)} {x.group(1)}", ir
+        r"<({.*})>\s*(\({[\S\s]*<{.*}>[^}]*}\))",
+        lambda x: f"{x.group(2)} {x.group(1)}",
```
jorendumoulin, line 22: L̷̹̒͆͜ô̵̻̼͂o̴̼̠̐̇k̸͔̍ś̶̮̭͛ ̸̨̛̝̔v̴͖̥́̅e̴̗͝r̷̠͗̚y̶̞͝ ̵͈̓́g̴̢̬̈́͠o̵̤͆o̸̦̭͠d̶͍͙̽!̴̓̀͜!̶̻͍͐

kernels/alloc/main.c
```diff
@@ -0,0 +1,50 @@
...
+
+int8_t *allocated_pointer;
+
+int8_t *_mlir_memref_to_llvm_alloc(uint32_t size) {
+  /* This calls malloc on the DMA core
+   * --> requires mlir opt to compile with:
+   *  --convert-memref-to-llvm="use-generic-functions index-bitwidth=32"
+   * To ensure that all cores in the cluster come up with the correct
+   */
+  snrt_cluster_hw_barrier();
```
jorendumoulin, line 17: This cluster synchronization is unnecessary?

jorendumoulin (changes requested):
Very nice! However, the test does not actually test the correctness of the allocation, I think.

kernels/alloc/main.c
JosseVanDelm: Good call :smile:

jorendumoulin (approved):
Perfect! :star_struck:

PR #29, opened 2023-11-23
memref.copy to snax dma

### LINKED ISSUE #30: Faulty packing C memref calling convention struct
There seems to be some problem when packing the memref C struct into an llvm struct.
The problem arises when investigating the shape of the struct:

```
struct OneDMemrefI32 {
  int32_t *data;
  int32_t *aligned_data;
  intptr_t  offset;
  intptr_t  shape[1];
  intptr_t  stride[1];
};
```

The shape is a pointer to the array of shapes of the memref.

Initializing such a 1d memref of size [10] looks like this:

```
  uint32_t constant_zero[1] = { 0 };
  uint32_t constant_one[1] = { 1 };
  uint32_t constant_size[1] = { 10 };

  // create memref object for A
  OneDMemrefI32_t memrefA = {
      .data = &A,
      .aligned_data = &A,
      .offset = constant_zero,
      .shape = constant_size,
      .stride = constant_one,
  };
```

When compiling the main.c to readable llvm the relevant lines look like this:
(compiling to readable llvm can be done by adding the following rule to `Makefile.rules`
```
%.ll: %.c
	$(CC) $(CFLAGS) -emit-llvm -S -c $< -o $@
```

Resulting llvm: (filtered relevant lines)
In this piece of code, llvm stores a pointer to the array, converted to an int in the struct.
```llvm
define dso_local i32 @main() local_unnamed_addr #0 {
  %3 = alloca [1 x i32], align 4   ; %3 = allocate the shape array for the resulting memref
  %4 = alloca %struct.OneDMemrefI32, align 4   ; %4 = the memref struct
  %11 = getelementptr inbounds [1 x i32], [1 x i32]* %3, i32 0, i32 0 ; %11 = the pointer to the first element of the shape array %3
  store i32 10, i32* %11, align 4    ; store 10 at the first element of the shape array
  ; get a pointer to the location of .shape in the memref struct:
  %17 = getelementptr inbounds %struct.OneDMemrefI32, %struct.OneDMemrefI32* %4, i32 0, i32 3, i32 0
  %18 = ptrtoint [1 x i32]* %3 to i32 ; convert the pointer to the shape array to i32
  store i32 %18, i32* %17, align 4, !tbaa !13 ; store the resulting int32 at the correct location in the struct
  ; execute a function using this struct:
  %36 = call i32 bitcast (i32 (...)* @_mlir_ciface_simple_copy to i32 (%struct.OneDMemrefI32*, %struct.OneDMemrefI32*)*)(%struct.OneDMemrefI32* nonnull %4, %struct.OneDMemrefI32* nonnull %5) #3
```

However, when extracting the shape information, MLIR seems to perform the following information: ([example](https://mlir.llvm.org/docs/TargetLLVMIR/#c-compatible-wrapper-emission)for a 2d matrix)
```mlir
  %4 = llvm.extractvalue %0[3, 0] : !llvm.memref_2d
  %5 = llvm.extractvalue %0[3, 1] : !llvm.memref_2d
```

The resulting value of `%4` and `%5` would then be pointers to the shape of their respective dimensions of the 2d matrix.
BUT, the result is of type index, not of type `index`, not of type `memref`. So MLIR expects these to be the actual shapes instead of pointers to the shapes?

An mlir op that extracts the dimension such as `memref.dim`:
```mlir
    %2 = "memref.dim"(%arg0, %0) : (memref<?xi32>, index) -> index
```
directly uses this value, instead of treating it as a pointer to the value.

Who is in the wrong here, the packing of the llvm structure or the unpacking. Is this the result of mixing LLVM versions? hmm

jorendumoulin:
This pull request translates `memref.copy` ops to snax dma transfers

Additionally, it fixes the faulty calling convention (resolves #30):
see https://discourse.llvm.org/t/using-c-compatible-memref-wrappers/75312/5


.github/workflows/python-tests.yml
jorendumoulin, line 1: this is just commenting out the entire test (just commenting out the "on" events still generates github notifications)

kernels/simple_copy/main.c
```diff
@@ -0,0 +1,60 @@
...
+#include "stdint.h"
+#include <snrt.h>
+
+// treat incoming size as pointer
+void _mlir_ciface_snax_dma_1d_transfer(size_t *source, size_t *destination,
+                                       size_t *size) {
+  snrt_dma_start_1d((void *)destination, (void *)source,
+                    *size * sizeof(size_t));
+  return;
+}
```
JosseVanDelm, line 12: Let's move this to a runtime library at some point.
Okay to keep it here now for LLD fun

kernels/simple_mult/main.c
```diff
@@ -71,6 +71,9 @@ int main() {
   _mlir_ciface_simple_mult(&memrefA, &memrefB, &memrefD);
   (void)snrt_mcycle();
 
+  static void *test;
+  static int test2;
+
   // Correctness check
   int nerr = 0;
   for (int i = 0; i < N; i++) {
```
JosseVanDelm, line 79: Why is this part of this PR?

kernels/simple_mult/test.c
```diff
@@ -0,0 +1,6 @@
+extern int *test;
+int main() {
+  int test2 = 34;
+  test = &test2;
+  return 0;
+}
```
JosseVanDelm, line 6: Why is this part of this PR?

tests/filecheck/transforms/copy_to_dma.mlir
```diff
@@ -0,0 +1,23 @@
+// RUN: ./compiler/snax-opt %s -p snax-copy-to-dma --print-op-generic | filecheck %s
+
+"builtin.module"() ({
+  "func.func"() <{"sym_name" = "simple_mult", "function_type" = (memref<?xi32>, memref<?xi32>) -> (), "sym_visibility" = "public"}> ({
+  ^0(%arg0 : memref<?xi32>, %arg1 : memref<?xi32>):
+    "memref.copy"(%arg0, %arg1) : (memref<?xi32>, memref<?xi32>) -> ()
+    "func.return"() : () -> ()
+  }) : () -> ()
+}) : () -> ()
```
JosseVanDelm, line 9: Can we also add a static test case and check whether mlir-opt can bypass the extract_aligned_pointer etc. immediately? I have the feeling these should be constant folded

kernels/simple_copy/main.c
```diff
@@ -0,0 +1,60 @@
+#include "data.h"
+#include "memref.h"
+#include "stdint.h"
+#include <snrt.h>
+
+// treat incoming size as pointer
+void _mlir_ciface_snax_dma_1d_transfer(size_t *source, size_t *destination,
+                                       size_t *size) {
+  snrt_dma_start_1d((void *)destination, (void *)source,
+                    *size * sizeof(size_t));
```
JosseVanDelm, line 10: Add a comment to a URL of your forum post

runtime/Makefile.rules
```diff
@@ -60,6 +60,9 @@ LDFLAGS += -lc
 LDFLAGS += -lsnRuntime
 LDFLAGS += $(SNITCH_RTL_PATH)/target/snitch_cluster/sw/snax/mac/build/mac.o
 
+%.ll: %.c
```
JosseVanDelm, line 63: This is only here for debugging purposes? Maybe add a comment?

compiler/transforms/snax_copy_to_dma.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class InsertFunctionCalls(RewritePattern):
+    """
+    Looks for memref copy operations and insert a snitch 1d dma call
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: memref.CopyOp, rewriter: PatternRewriter):
+        # Exctract size information
```
JosseVanDelm, line 20:
```suggestion
        # Extract size information
```

compiler/transforms/snax_copy_to_dma.py
```diff
@@ -0,0 +1,61 @@
...
+            "snax_dma_1d_transfer",
+            [source_ptr_op.aligned_pointer, dest_ptr_op.aligned_pointer, dim_op.result],
+            [],
+        )
+
+        # Replace op with function call
+        rewriter.insert_op_before_matched_op(
+            [zero_const, dim_op, source_ptr_op, dest_ptr_op]
+        )
+        rewriter.replace_op(op, func_call)
```
JosseVanDelm, line 39: Wondering whether we should check here if the memref.copy is actually copying a 1D memref?
Otherwise some reshaping is necessary?

kernels/simple_copy/baseline.c
JosseVanDelm, line 1: Maybe rename this file? This is not really a baseline.
The baseline was for experiments in the other repo where MLIR emitted everything, not C code.

compiler/transforms/snax_copy_to_dma.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class SNAXCopyToDMA(ModulePass):
```
JosseVanDelm, line 42: Maybe convert-memref-to-snax sounds a bit more idiomatic?
The main thing that I don't like about this naming is that it's not really a SNAX to something, it's more a something to snax.

tests/filecheck/transforms/copy_to_dma.mlir
```diff
@@ -0,0 +1,23 @@
...
+
+//CHECK: "builtin.module"() ({
+//CHECK-NEXT:   "func.func"() <{"sym_name" = "simple_mult", "function_type" = (memref<?xi32>, memref<?xi32>) -> (), "sym_visibility" = "public"}> ({
+//CHECK-NEXT:   ^0(%arg0 : memref<?xi32>, %arg1 : memref<?xi32>):
```
JosseVanDelm, line 13: Does this still work for 2D memrefs?
Should it fail if it encounters such memrefs?

JosseVanDelm (changes requested):
Very nice! Please see my comments!
Let's also include a link to this discussion:
https://discourse.llvm.org/t/using-c-compatible-memref-wrappers/75312

kernels/simple_mult/main.c
jorendumoulin: by accident 😯 
sorry for this

tests/filecheck/transforms/copy_to_dma.mlir
jorendumoulin: Yes, but this happens during inlining or at the LLVM level, so not really a test for this transform?

kernels/simple_copy/baseline.c
jorendumoulin: In fact, it is a stupid check alltogether, so I removed the file

compiler/transforms/snax_copy_to_dma.py
jorendumoulin: Yes, reshaping is not supported yet, so I just included a check that both memrefs are of rank 1

compiler/transforms/snax_copy_to_dma.py
jorendumoulin: On the other hand, I think its maybe more clear to let all of our custom transforms part of the SNAX platform to start with 'snax' so it is very clear which passes are custom and which are not?

something like this: https://xilinx.github.io/mlir-air/AIRTransformPasses.html

Also most mlir passes seem to start with the dialect they belong to (there are exceptions though)

tests/filecheck/transforms/copy_to_dma.mlir
jorendumoulin: with the new check,  it ignores memrefs with more than 1 dimension for now

PR #31, opened 2023-12-04
Compiler: add transform snax-to-func

jorendumoulin:
This pass transform snax dialect ops to external function calls, such that they can be implemented with the C runtime of snitch

For now, this is only the cluster sync op


JosseVanDelm (approved):
Nice!

PR #35, opened 2023-12-04
simple_mult end to end

### LINKED ISSUE #32: Solve address space issues
The generated address spaces in MLIR cause some error further down the pipeline on the following lines:

```
    %28 = llvm.call @_mlir_memref_to_llvm_alloc(%27) : (i32) -> !llvm.ptr<i8, 1>
    %29 = llvm.bitcast %28 : !llvm.ptr<i8, 1> to !llvm.ptr<i32, 1>
```

A bitcast cannot convert between address spaces.
Because all conversions between address spaces have all been handled in previous compiler passes (see #16 ), the address space information is no longer relevant, we therefore consider just creating a pass to remove all this information (`clear-memory-spaces`), and avoid these kinds of errors down the pipeline

jorendumoulin:
Complete the end to end workflow for the simple_mult kernel

This includes the following fixes
- Include entire mlir flow in the makefile, now all memory management and syncrhonization is handled by the mlir transformation passes 🥳 
- Clear address spaces before emitting llvm (resolves #32 )
- Some renaming of functions to avoid duplicate names between mlir <-> snitch runtime
- Creation of a snax runtime library with an implementation of the functions called from mlir
- Fix for regex with duplicate attributes (`op {attr 1} {attr 2} : ...` -> `op {attr 1, attr 2} :`) (the duplicate attributes are a result of previous regex transformations.


runtime/include/snax_rt.h
```diff
@@ -0,0 +1,36 @@
...
+                                       size_t size) {
+  snrt_dma_start_1d((void *)destination, (void *)source, size * sizeof(size_t));
+  return;
+}
+
+int8_t _mlir_ciface_snax_is_dm_core() { return (int8_t)snrt_is_dm_core(); }
+
+int8_t _mlir_ciface_snax_is_compute_core() {
+  return (int8_t)snrt_is_compute_core();
+}
```
JosseVanDelm, line 36: Why do the cast to `int8_t`? Is this the way the ciface expects it?

kernels/simple_mult/main.c
JosseVanDelm, line 1: Shouldn't you add the dma_transfer inside the mlir file here?
Now you seem to be accessing L3 memory directly :upside_down_face:

compiler/transforms/clear_memory_space.py
```diff
@@ -0,0 +1,35 @@
...
+            return t
+
+        for op_in_module in module.walk():
+            for operand in op_in_module.operands:
+                operand.type = clear_memory_space(operand.type)
+
+            # special case for func ops
+            if isinstance(op_in_module, func.FuncOp):
+                # Define new function type with updated inputs and outputs
+                # mapped to a default memory space
```
JosseVanDelm, line 29: I would put these comments together :smile:

JosseVanDelm (changes requested):
~Are you sure the copy to L1 is performed?~

It appears I'm a dumbdumb

kernels/simple_mult/main.c
JosseVanDelm: @JosseVanDelm you silly fool it already inserts all of these things automagically

JosseVanDelm (approved):
(Woops) very cool!

runtime/include/snax_rt.h
jorendumoulin: changed to use int

PR #36, opened 2023-12-04
Build snax-mac and snax-gemm in one dockerfile

JosseVanDelm:
This adds:
* A new base container/digest from dockerhub that also contains Chisel dependencies
* A common build step that installs common dependencies
* Extra separate build stage for both RTLs and copies them in a new folder
* A new commit from the snitch_cluster RTL repo that includes both configurations of SNAX (snax-mac and snax-gemm). This removes:
* The config argument for a docker build, all configs are now built inside one container


runtime/Makefile.rules
```diff
@@ -4,12 +4,12 @@ MAKEFILE_RULES_DIRNAME := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
...
 # NOTE: we are using  absolute paths from the Docker container
 SNITCH_LLVM_PATH = /opt/snitch-llvm
-SNITCH_RTL_PATH = /opt/snitch_cluster
+SNITCH_SW_PATH = /opt/snax-mac
 CC            = $(SNITCH_LLVM_PATH)/bin/riscv32-unknown-elf-clang
 LD            = $(SNITCH_LLVM_PATH)/bin/riscv32-unknown-elf-clang
 AR            = $(SNITCH_LLVM_PATH)/bin/llvm-ar
 RANLIB        = $(SNITCH_LLVM_PATH)/bin/llvm-ranlib
-VLTSIM        = /opt/snitch-rtl/bin/snitch_cluster.vlt
+VLTSIM        = /opt/snax-mac-rtl/bin/snitch_cluster.vlt
```
jorendumoulin, line 12: If the correct `VLTSIM` is now selected by the kernel, does it make sense to define this parameter seperately in every kernel `Makefile`, instead of the common one? I don't think it is used in the common `Makefile.rules` anyway.

jorendumoulin (approved):
Nice 😄 
If you can run the tests again on the new container build, this is good for me!
I have one small suggestion:

runtime/Makefile.rules
JosseVanDelm: @jorendumoulin How do you feel about #38 ?

jorendumoulin: #38 is great!

PR #40, opened 2023-12-07
Move dependences to requirements.txt

tobiasgrosser:
Filecheck and lit are already moved, so we can just drop them.


JosseVanDelm:
@tobiasgrosser I know what it is about! Will resolve later :) Our CI pipeline wrt docker is a bit flaky :>

PR #41, opened 2023-12-07
compiler: extend kernel matching

jorendumoulin:
This PR uses #68 (from the future 🤯) to update the mult kernel matching and also implements the matmul kernel matching.

This is done based on the combination of kernel types and dataflows.

For mult,
We expect an elementwise 1d dataflow, and a MUL kernel.

For matmul
We expect a 3d dataflow with 1 reducing dimension, and a MAC kernel

Both dispatchments do not care about data layout for now, that mechanism is for a future PR!


JosseVanDelm:
Cool!

runtime/Makefile.rules
```diff
@@ -99,7 +99,7 @@ MLIRPREPROC2FLAGS += --mlir-print-local-scope
 
 # SNAX opt
 
-SNAXOPTFLAGS = -p set-memory-space,insert-sync-barrier,dispatch-regions,dispatch-elementwise-mult,linalg-to-library-call,snax-copy-to-dma,snax-to-func,clear-memory-space
+SNAXOPTFLAGS = -p set-memory-space,insert-sync-barrier,dispatch-regions,dispatch-kernels,linalg-to-library-call,snax-copy-to-dma,snax-to-func,clear-memory-space
```
JosseVanDelm, line 102: Maybe it's time to start splitting this line :exploding_head: 
```suggestion
SNAXOPTFLAGS = -p 
SNAXOPTFLAGS +=set-memory-space,
SNAXOPTFLAGS +=insert-sync-barrier,
SNAXOPTFLAGS +=dispatch-regions,
SNAXOPTFLAGS +=dispatch-kernels,
SNAXOPTFLAGS +=linalg-to-library-call,
SNAXOPTFLAGS +=snax-copy-to-dma,
SNAXOPTFLAGS +=snax-to-func,
SNAXOPTFLAGS +=clear-memory-space
```

compiler/transforms/dispatch_kernels.py
```diff
@@ -0,0 +1,133 @@
...
+                return
+
+            if not isinstance(inp.type.get_element_type(), builtin.IntegerType):
+                return
+
+        if len(op.iterator_types) != 1:
+            return
+
+        if op.iterator_types.data[0].data is not linalg.IteratorType.PARALLEL:
+            return
```
JosseVanDelm, line 50: Not sure if we can reorder the for loop out of this, but again I think we can fuse some of this stuff together?
Let me know if you don't like this style!
```suggestion
        if any( 
            op.library_call is not None,
            KernelType.get_kernel(op) != KernelType.MUL,
            ...
         ):
            return
        for inp in op.inputs:
             if any(
                         ...
                    ):
                    return
       # Move these up?
       # if len(op.iterator_types) != 1:
       #     return

       # if op.iterator_types.data[0].data is not linalg.IteratorType.PARALLEL:
       #     return
```

compiler/transforms/dispatch_kernels.py
```diff
@@ -0,0 +1,133 @@
...
+
+        for inp in [x for x in op.inputs if isinstance(x.type, builtin.ShapedType)]:
+            if not isinstance(inp.type, MemRefType):
+                return
+            if len(inp.type.get_shape()) != 2:
+                return
+            if not isinstance(inp.type.get_element_type(), builtin.IntegerType):
+                return
+
```
JosseVanDelm, line 81: See again previous comment, also on previous lines

compiler/transforms/dispatch_kernels.py
JosseVanDelm: On second thought, I think applying my suggestions messes up performance :( :
https://stackoverflow.com/a/53600865/18472362

compiler/transforms/dispatch_kernels.py
JosseVanDelm: Damnit I messed up, see my second comment to my previous comment

JosseVanDelm (approved):
This looks very cool! Although I think it seems a bit arbitrary what logic you are splitting up in KernelType enum and which "logic" you actually add to these passes?

PR #45, opened 2023-12-12
add dynamic alloc support

jorendumoulin:
This PR adds dynamic shape support for the pass `set-memory-spaces`. 
If the pass realizes a memory cast on dynamically shaped memrefs, the dynamic allocations are now correctly handled.

The kernel `simple_mult` is adapted to use dynamic shapes.

### DIFF
--- a/compiler/transforms/set_memory_space.py
+++ b/compiler/transforms/set_memory_space.py
@@ -1,4 +1,4 @@
-from xdsl.dialects import func, builtin, memref, linalg
+from xdsl.dialects import func, builtin, memref, linalg, arith
 from xdsl.ir import MLContext
 from xdsl.ir.core import SSAValue
 from xdsl.passes import ModulePass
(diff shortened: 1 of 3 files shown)


compiler/transforms/set_memory_space.py
```diff
@@ -155,13 +155,25 @@ def match_and_rewrite(self, op: memref.MemorySpaceCast, rewriter: PatternRewrite
         if not isinstance(op.results[0].type, memref.MemRefType):
             return
 
+        # create memref.dim operations for dynamic dimensions
+        shapes = [x.value.data for x in op.results[0].type.shape.data]
+        dyn_operands = []
+        for i in range(len(shapes)):
+            if shapes[i] == -1:
```
JosseVanDelm, line 162:
```suggestion
            # Dynamic shapes are represented as -1
            if shapes[i] == -1:
```

JosseVanDelm (approved):
Nice!

Issue #48, opened 2023-12-21
Tracking issue: Move to generic runtime


JosseVanDelm:
Since KULeuven-MICAS/snitch_cluster#57 we have a generic LLVM support to move to LLVM main.

This would allow us to:
* Remove opaque pointer support (and ditch `mlir-translate-16`)
* Use properties in MLIR (and ditch `tomlir16.py` and `tomlir18.py`)
* Have access to new bytecode support (needed for TOSA import from tensorflow)

To complete this issue, I need to:
- [x] Update the version in the container of KULeuven-MICAS/snitch_cluster to use the version with the generic runtime
  - [x] Adapt sha's and commit hashes
  - [x] Install llvm
  - [x] Add rtl-generic paths
- [x] Change the rules of compilation in `runtime/Makefile.rules` to:
  - [x] Use the generic riscv target
  - [x] Not use tomlir-16/18.py anymore
  - [x] Use a different version of LLVM
- [ ] Update readme to reflect newest flow

PR #46, opened 2023-12-19
bump xdsl

jorendumoulin:
Bump xDSL to get the nice PR xdslproject/xdsl#1825

PR #49, opened 2023-12-21
Use upstream LLVM with generic runtime

JosseVanDelm:
Solves #48

Note that there are still two linking dependencies that I'm hauling over from the old toolchain.
In KULeuven-MICAS/snitch_cluster#57 this toolchain was installed by default, here I have to explicitly keep it in the installation but also copy over the static libraries.
* `libclang_rt.builtins-riscv32.a`: this is not built with the standard distribution of LLVM and I'm too lazy to build this myself.
More info here (I think) : https://github.com/llvm/llvm-project/tree/main/compiler-rt/lib/builtins
* `libc.a` apparently there are a few routines in the runtime that rely on `memset`: Not sure why we would have to stick to this specific version, but hey it works. Compiling directly with only the source of memset might save us a few kb's in the final elf file size, but I'm not sure about this.
One implementation is in musl: https://git.musl-libc.org/cgit/musl/tree/src/string/memset.c


runtime/tomlir18.py
jorendumoulin, line 1: 😄

runtime/tomlir16.py
jorendumoulin, line 1: 😄

jorendumoulin (approved):

JosseVanDelm:
Using the files and building locally,
Undoing #46

I have these issues:
### Simple copy :x:
```
 main.c:28:5: warning: implicit declaration of function '_mlir_ciface_simple_copy' is invalid in C99 [-Wimplicit-function-declaration]
    _mlir_ciface_simple_copy(&memrefA, &memrefB);
    ^ 
```
In clang 17 this is an error instead of a warning, and the function has to be declared before it can be used.
```patch
diff --git a/kernels/simple_copy/main.c b/kernels/simple_copy/main.c
index eeed3fc..0e1fa5a 100644
--- a/kernels/simple_copy/main.c
+++ b/kernels/simple_copy/main.c
@@ -5,6 +5,9 @@
 #include <snrt.h>
 #include <stdint.h>
 
+
+void _mlir_ciface_simple_copy(OneDMemrefI32_t* memrefA, OneDMemrefI32_t* memrefB);
+
 int main() {
 
   // create memref object for A
```
Furthermore, it fails:
```
Warning: Failed to write binary name to logs/.rtlbinary
[fesvr] Wrote 36 bytes of bootrom to 0x1000
[fesvr] Wrote entry point 0x80000000 to bootloader slot 0x1020
[fesvr] Wrote 56 bytes of bootdata to 0x1024
[Tracer] Logging Hart          0 to logs/trace_hart_00000000.dasm
[Tracer] Logging Hart          1 to logs/trace_hart_00000001.dasm
*** FAILED *** (tohost = 1)
```
### Simple mult :heavy_check_mark: 

### Simple matmult :heavy_check_mark: 

After applying this fix:
```patch
diff --git a/kernels/simple_matmul/main.c b/kernels/simple_matmul/main.c
index 1e37e3c..00c7789 100644
--- a/kernels/simple_matmul/main.c
+++ b/kernels/simple_matmul/main.c
@@ -49,7 +49,7 @@ void _mlir_ciface_simple_matmul_cpu(TwoDMemrefI8_t *a, TwoDMemrefI8_t *b,
   int8_t *a_ptr = a->aligned_data;
   int8_t *b_ptr = b->aligned_data;
   int32_t *c_ptr = c->aligned_data;
-  batch_gemm_cpu(Batch, M_param, K_param, N_param, a_ptr, b_ptr, c_ptr,
+  batch_gemm_cpu(Batch, M_param, K_param, N_param, a_ptr, b_ptr, 0, 0, c_ptr,
                  strideInnermostA, strideInnermostB, strideInnermostC, ldA, ldB,
                  ldC, strideA, strideB, strideC);
 }
```
## Alloc :heavy_check_mark:

jorendumoulin:
I fixed the `simple_copy` kernel! See #54 
I also fixed the xdsl thingy, see #53

JosseVanDelm:
Awaiting merging of #53

PR #53, opened 2024-01-03
fix xdsl bug

### LINKED ISSUE #52: New xdsl version was not properly tested by CI?
It looks like using the docker cache with the newer version of xdsl is breaking some things:
```
/repo/runtime//../compiler/snax-opt -p set-memory-space,insert-sync-barrier,dispatch-regions,dispatch-elementwise-mult,linalg-to-library-call,snax-copy-to-dma,snax-to-func,clear-memory-space -o linalg.snax-opt.mlir linalg.preproc.mlir
/opt/python3.11/lib/python3.11/site-packages/xdsl/utils/deprecation.py:18: UserWarning: Call to deprecated method MemRefType.from_element_type_and_shape: use the constructor (`ClassName(...)`) instead.
  warnings.warn(
Traceback (most recent call last):
  File "/repo/runtime//../compiler/snax-opt", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/repo/compiler/tools/snax_opt_main.py", line 58, in main
    SNAXOptMain().run()
  File "/opt/python3.11/lib/python3.11/site-packages/xdsl/xdsl_opt_main.py", line 70, in run
    if self.apply_passes(module):
       ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/python3.11/lib/python3.11/site-packages/xdsl/xdsl_opt_main.py", line 276, in apply_passes
    self.pipeline.apply(self.ctx, prog)
  File "/opt/python3.11/lib/python3.11/site-packages/xdsl/passes.py", line 177, in apply
    prev.apply(ctx, op)
  File "/repo/compiler/transforms/set_memory_space.py", line 229, in apply
    PatternRewriteWalker(RealizeMemorySpaceCasts()).rewrite_module(op)
  File "/opt/python3.11/lib/python3.11/site-packages/xdsl/pattern_rewriter.py", line 641, in rewrite_module
    self._rewrite_op(op)
  File "/opt/python3.11/lib/python3.11/site-packages/xdsl/pattern_rewriter.py", line 689, in _rewrite_op
    self._rewrite_op_regions(op)
  File "/opt/python3.11/lib/python3.11/site-packages/xdsl/pattern_rewriter.py", line 702, in _rewrite_op_regions
    iter_op = self._rewrite_op(iter_op)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/python3.11/lib/python3.11/site-packages/xdsl/pattern_rewriter.py", line 689, in _rewrite_op
    self._rewrite_op_regions(op)
  File "/opt/python3.11/lib/python3.11/site-packages/xdsl/pattern_rewriter.py", line 702, in _rewrite_op_regions
    iter_op = self._rewrite_op(iter_op)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/python3.11/lib/python3.11/site-packages/xdsl/pattern_rewriter.py", line 657, in _rewrite_op
    self.pattern.match_and_rewrite(op, rewriter)
  File "/opt/python3.11/lib/python3.11/site-packages/xdsl/pattern_rewriter.py", line 454, in impl
    func(self, op, rewriter)
  File "/repo/compiler/transforms/set_memory_space.py", line 158, in match_and_rewrite
    shapes = [x.value.data for x in op.results[0].type.shape.data]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/repo/compiler/transforms/set_memory_space.py", line 158, in <listcomp>
    shapes = [x.value.data for x in op.results[0].type.shape.data]
              ^^^^^^^
AttributeError: 'IntAttr' object has no attribute 'value'
../../runtime/Makefile.rules:95: recipe for target 'linalg.snax-opt.mlir' failed
```
Is this related to https://github.com/xdslproject/xdsl/pull/1825

jorendumoulin:
This PR aims to resolve #52, partly.
When bumping xdsl, the new version is only installed in the docker build test, but other python tests still run on the docker build from the main branch.

I included rules to first update the python requirements from requirements.txt for python tests, such that the correct version of xdsl is used.

This resolves the problem when bumping modules defined in requirements.txt. However, this does not solve the problem for modules only defined in the Dockerfile (such as the snax-cluster).

Additionaly, this also includes the bugfix of the changed IntAttr xdslproject/xdsl#1825

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Note: leaving out the `config` `--build-arg` will use the default `snitch_cluste
 
 Our github actions use the container build to run tests. The actions always use the last docker build from the `main` branch. Be aware that if you make changes to the Dockerfile in a certain PR, the actions will still run on the `main` Docker build, which may result in errors. Changes to `requirements.txt` are reinstalled before every test, so these are safe to change. The following packages are only defined in the Dockerfile and thus require extra caution when changing:
 
-* [snax-cluster](https://github.com/kuleuven-micas/snitch_cluster) and dependencies (runtime, verilator, runtime, spike)
+* [snax-cluster](https://github.com/kuleuven-micas/snitch_cluster) and dependencies (runtime, verilator, spike)
 * llvm (llvm, clang, lld)
 * standard linux packages
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Our github actions use the container build to run tests. The actions always use 
 
 * [snax-cluster](https://github.com/kuleuven-micas/snitch_cluster) and dependencies (runtime, verilator, spike)
 * llvm (llvm, clang, lld)
-* standard linux packages
+* standard linux packages (everything installed with `apt`)
 
 ## Pytorch -> Linalg
 
--- a/README.md
+++ b/README.md
@@ -13,15 +13,17 @@ pytorch -> `torch-mlir` -> linalg dialect -> mlir-opt, mlir-translate, clang -> 
 
 ## Setup with Docker Container
 
-For compiling the low-level kernels you need the snitch compilation toolchain, 
+For compiling the low-level kernels you need the snitch compilation toolchain,
 which is easiest to get through a docker container.
 
 ### Getting the container remotely
 
 You can run tests/experiments in this repository with:
+
 ```sh
 docker run -itv `pwd`:/repo:z ghcr.io/kuleuven-micas/snax-mlir:main
 ```
+
 This will download the image if it is not present on your system yet.
 The repository will be available under `/repo`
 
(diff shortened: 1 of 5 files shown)


JosseVanDelm (changes requested):
Good, but I'd like to add a line to the readme for users of the docker to also perform the reinstallation, otherwise, this might lead to confusion and difficult to debug issues for external users.

Some other thoughts, though maybe more relevant for later:

An other option is to install the packages on every run and not install it in the docker in  the first place. 
The python dependencies are most likely to be the fastest moving targets anyway, so it kind of makes sense to remove them. On the other hand, if we would have to install tensorflow on each run, this might significantly slow down our testing....
I also think github has a setup-python action that can cache python dependencies for you, but then we might run into caching issues :smile: 
  
What you are suggesting here is probably faster, but might be a bit confusing to people not familiar with our testing pipeline, altough, that seems less important in this stage of the project.

jorendumoulin:
Agreed! I updated the readme. What we have now though is I think the best of both worlds: we both install the packages in the docker build **and** install them in the Docker build. If they change, pip will reinstall them. If they don't change, they are pre-installed and the test runs quickly!

README.md
```diff
@@ -34,15 +36,25 @@ pip3 install -e /repo
...
 
 Note: leaving out the `config` `--build-arg` will use the default `snitch_cluster` setup.
 
+### A note on github actions
+
+Our github actions use the container build to run tests. The actions always use the last docker build from the `main` branch. Be aware that if you make changes to the Dockerfile in a certain PR, the actions will still run on the `main` Docker build, which may result in errors. Changes to `requirements.txt` are reinstalled before every test, so these are safe to change. The following packages are only defined in the Dockerfile and thus require extra caution when changing:
+
+* [snax-cluster](https://github.com/kuleuven-micas/snitch_cluster) and dependencies (runtime, verilator, spike)
+* llvm (llvm, clang, lld)
+* standard linux packages
```
JosseVanDelm, line 56:
```suggestion
* standard linux packages (everything installed with `apt`)
```

JosseVanDelm (approved):

PR #54, opened 2024-01-03
fix simple_copy kernel

jorendumoulin:
This PR fixes the `simple_copy` kernel for LLVM17.

Apperantly, the turbo-charged blazingly quick llvm17 code is so fast that the kernel starts checking the result before the DMA transfer is finished. For this reason, i suggest simply inserting a dma wait by default in the runtime. This can later be optimized such that we only wait for DMA results when we need them, but no system like that is in place right now.

### DIFF
--- a/kernels/simple_copy/main.c
+++ b/kernels/simple_copy/main.c
@@ -5,6 +5,9 @@
 #include <snrt.h>
 #include <stdint.h>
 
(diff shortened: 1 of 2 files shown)


JosseVanDelm (approved):
Thank you!!!

PR #56, opened 2024-01-08
Manual action triggers

jorendumoulin:
This PR was an attempt to better structure the github action tests.

The goal was to run the actual tests (such as build and run kernels), based on the `build-docker` action
* If build docker is skipped, test right away using the `main` docker build
* If build docker build is run, run the tests after the docker build is finished, with the container tag of the current branch (pr-xx) or main.

I thought this would be possible using the following mechanisms:

1. using separate jobs and sequential executing with `needs`: this may be possible, but then all the jobs and workflows are in the same file, and the docker container is built even when not necessary.
2. using `workflow_run`: this trigger runs a workflow if another workflow is finished, but for some reason this is only possible on the main branch.

Instead of this, I just made manual workflow triggers which can be activated in the actions tab:
manual_workflow

(The manual trigger will be available as soon as this PR is merged into main)

The automatic workflows still use the `main` docker build as before. But, if you want to run the actions on a newly built docker container, you can by manually entering the correct container tag (for example `pr-56`).


JosseVanDelm (approved):
wicked

PR #58, opened 2024-01-11
Alter actions pull request trigger

jorendumoulin:
Also run actions on pull requests that do not go to the main branch. I want this for #55 for example, and there is no reason to not do this.


JosseVanDelm (approved):
Sure!

PR #63, opened 2024-01-22
compiler: transform copies

jorendumoulin:
This PR implements the layout transformations between different TSL memory layouts.

The layout transformations are included in `snax-copy-to-dma.py`. The previously existing pass is renamed to `MatchSimpleCopy`, which resolves memory copies between two memrefs without layout transformation. Additionally, `TransformDMA` handles memory copies where both memrefs have a `TSL` memory layout attribute.

The `TransformDMA` generates a nested set of for loops to correctly execute the layout transformation. For this, we need ops for the bounds and steps of the Strides in the TSL. `tsl.py` is extended with two functions to generate bound ops and step ops for a given TSL. Given that a TSL has some degrees of dynamic values, this is somewhat complicated.

With this PR, the `transform_copy` kernel is also added. A simple example is present in `transform_copy.preproc.mlir`. Additionally, to quickly generate a lot of different tests and run experiments, the kernel also includes a test generation framework. Based on a description in `testcases.py`, the `make generate` command will generate different MLIR files and folders with the correct `data.c` and `data.h` files. The test will then do a transform copy, and check if the resulting data is the same as the golden model generated by `gentests.py`.


compiler/transforms/snax_copy_to_dma.py
```diff
@@ -9,47 +12,322 @@
...
+        for i in range(len(remaining_strides)):
+            next_for_op = for_loop.body.block.first_op
+            # insert the ops in the for loop body
+            ops_to_insert_for_loop = []
+
+            # source indexing operations:
+            # stride_src = Constant.from_int_and_width(
+            #     remaining_strides[i][0].step, IndexType()
+            # )
+            stride_src = remaining_strides[i]["step_src_op"]
```
JosseVanDelm, line 286: What is the purpose of this comment?

JosseVanDelm (changes requested):
Very nice PR!

Way too big though :angry: :wink:

The test style can be improved a bit in my opinion, maybe in a later PR? let's discuss offline

jorendumoulin:
I've udpated the MLIR test generation to use the implicit builder and applied all other minor suggestions! For a better code refactoring of all tests, i made #67

JosseVanDelm (approved):
Very nice work @jorendumoulin !

PR #65, opened 2024-01-23
fix tsl parsing error

jorendumoulin:
The TSL still had a reverse operation when parsing the strides and bounds, interpreting the innermost strides as the outermost ones. The filecheck tests did not catch this error because the XDSL roundtrip parses and prints the IR twice, putting the steps and bounds back in their correct order 😒 

Additionaly i also renamed some functions, to be more consistent with the naming of:
A `Stride` consists of a `step` and a `bound`

### DIFF
--- a/compiler/parser/tsl_parser.py
+++ b/compiler/parser/tsl_parser.py
@@ -19,16 +19,16 @@ class TSLParser(BaseParser):
             return v
         self.raise_error("Expected an integer literal or `?`" + context_msg)
 
-    def _parse_stride(self) -> list[int]:
+    def _parse_step(self) -> list[int]:
         """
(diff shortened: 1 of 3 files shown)


JosseVanDelm (approved):
Small nit :)

> The filecheck tests did not catch this error because the XDSL roundtrip parses and prints the IR twice, putting the steps and bounds back in their correct order 😒

Also yikes, should we add a test that does a single trip (would that be how they call it, lol?) ?

jorendumoulin:
Thanks! I added a _singletrip_ test (1 pass), which can be used in parallel to the roundtrip test (2 passes).
Maybe in the future we can add _randomtrip_ which does it a random amount of times haha

jorendumoulin:
trippy 🌃 🍄

PR #68, opened 2024-01-29
add kernel_type util

jorendumoulin:
This PR inits the kernel_type util

The Kernel Type is an enum, to classify different kernels of a linalg body. For now, the code can classify the following kernels:

- MUL
- MAC
- QMAC

however, extending for other types of kernels should be easy with the implemented structure.

These changes pave the way for linalg.generic matching.

The Enum has a lot of functions. The most important one is `get_kernel`, where we identify the kernel type based on linalg operation as input. All of the other functions are helper parser functions to perform this identification.

### DIFF
--- a/compiler/util/kernel_type.py
+++ b/compiler/util/kernel_type.py
@@ -25,6 +25,19 @@ class InputException(KernelException):
 
 
 class KernelType(Enum):
+    """
+    Enumeration representing different types of kernel operations of a
+    linalg generic body.
+
+    Attributes:
+        MUL (str): Represents a multiplication operation.
+            out = a * b
+        MAC (str): Represents a multiply-accumulate operation.
+            out += a * b
+        QMAC (str): Represents a quantized multiply-accumulate operation.
+            out += (a - zp_a) * (b - zp_b)
+    """
+
     MUL = "mul"
     MAC = "mac"
     QMAC = "qmac"
@@ -201,6 +214,8 @@ class KernelType(Enum):
             return None
 
         # make sure yield_op is produced by other op before parsing
+        # if not, the kernel is empty (just linalg yield), and
+        # yielded_op.op would result in an AttributeError
         yielded_op = yield_op.operands[0]
         if not isinstance(yielded_op, OpResult):
             return None
--- a/compiler/util/kernel_type.py
+++ b/compiler/util/kernel_type.py
@@ -6,10 +6,21 @@ from xdsl.ir import Block, Operation, OpResult, SSAValue
 
 
 class KernelException(Exception):
+    """
+    A KernelException is raised by the parsing functions if the provided
+    operation does not match the expected operation type in the kernel.
+    """
+
     pass
 
 
 class InputException(KernelException):
+    """
+    A InputException is a specific case of parsing error, when we fail to
+    match given operands to the inputs of the kernel. The current parsing
+    strategy is therefore not necessarily incorrect, just incomplete.
+    """
+
     pass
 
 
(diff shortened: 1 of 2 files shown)


compiler/util/kernel_type.py
```diff
@@ -0,0 +1,240 @@
...
+        Args:
+            linalg_op (linalg.Generic): The linalg operation.
+
+        Returns:
+            tuple[linalg.YieldOp, dict]: The yield operation and a
+            dictionary of input types.
+
+        Raises:
+            KernelException: If the number of shaped inputs is incorrect for binary ops.
+            KernelException: If the last op of the linalg block is not a yield op.
```
JosseVanDelm, line 122: I don't think we need to check this here.
To me it seems like this is already checked by the verifier, each `linalg.generic` must be terminated by a `linalg.yield`, since it has the `SingleBlockImplicitTerminator<YieldOp>` trait :)

compiler/util/kernel_type.py
```diff
@@ -0,0 +1,240 @@
...
+            KernelException: If the yield op does not return exactly one result.
+            KernelException: If the yield op operand is not the result of another op.
+        """
+
+        linalg_block: Block = linalg_op.body.block
+        input_types: list = [o.type for o in linalg_op.operands]
```
JosseVanDelm, line 128: These are also the `outs`, right? I find this naming a bit confusing, maybe change to `operands`

compiler/util/kernel_type.py
```diff
@@ -0,0 +1,240 @@
...
+        # block are tensors or scalars, therefore we generate a dictionary
+        # of the input types of the linalg
+        types = dict(zip(linalg_block.args, input_types))
+
+        # this check counts the number of inputs which are shaped inputs
+        # for all the kernels there may be three shaped operands
+        # (input a, input b, output c)  other inputs may be integer types used for
+        # constant values, such as zero-point offsets
+        if len([type for type in types.values() if isinstance(type, ShapedType)]) != 3:
+            raise KernelException("Wrong number of shaped inputs")
```
JosseVanDelm, line 141: Maybe make this more explicit?
```suggestion
            raise KernelException("Wrong number of shaped inputs: Expect 2 shaped inputs and 1 shaped output")
```

compiler/util/kernel_type.py
```diff
@@ -0,0 +1,240 @@
...
+
+        # last op of the linalg block should be a yield op
+        yield_op = linalg_block.last_op
+        if not isinstance(yield_op, linalg.YieldOp):
+            raise KernelException("Last op of linalg block is not a yield op")
+        # yield op should have one result, which is produced by an op
+        if len(yield_op.operands) != 1:
+            raise KernelException("Yield op does not have one operand")
```
JosseVanDelm, line 149: This check doesn't make a lot of sense to me.
Can't we just check if `len(linalg_op.outputs) == 1`?

compiler/util/kernel_type.py
```diff
@@ -0,0 +1,240 @@
...
+
+        # input parsing successful, we can return the computed types
+        # for further checking
+        return (yield_op, types)
+
+    @staticmethod
+    def match_inputs(a: Operation, b: Operation, types: dict):
+        """
+        Matches the operands a and b with the input types of the linalg kernel.
+        This is mainly used as a check to see if we have reachted the top
```
JosseVanDelm, line 161:
```suggestion
        This is mainly used as a check to see if we have reached the top
```

JosseVanDelm (commented):
I'm not sure I fully understand what we are doing here.
It seems like some of the checks are superfluous because they are already verified.
Or like the could be implemented in a more straightforward fashion?
Can we discuss offline?

compiler/util/kernel_type.py
```diff
@@ -197,10 +200,15 @@ def get_kernel(linalg_op: linalg.Generic):
         except KernelException:
             return None
 
+        # make sure yield_op is produced by other op before parsing
+        yielded_op = yield_op.operands[0]
+        if not isinstance(yielded_op, OpResult):
+            return None
+
```
JosseVanDelm, line 207: I still don't understand why this check is necessary, can you add a comment?

compiler/util/kernel_type.py
```diff
@@ -0,0 +1,248 @@
...
+    strategy is therefore not necessarily incorrect, just incomplete.
+    """
+
+    pass
+
+
+class KernelType(Enum):
+    MUL = "mul"
+    MAC = "mac"
+    QMAC = "qmac"
```
JosseVanDelm, line 30: Add a docstring to this class, and a short comment on what each specific type is?
I don't remember exactly what you mean with a qmac and a mul operation tbh :shrug:

JosseVanDelm (commented):
Can you add some more comments?

JosseVanDelm (approved):

PR #69, opened 2024-02-01
mlperf tiny: generalize tensor.pad operations

JosseVanDelm:
Not using a generalization pass to convert `tensor.pad` into `linalg.generic`s in our current pipeline yields "weird" `linalg.map`, which is apparently not what we want. This PR adds a new pass to circumvent that problem, and also uses a newer version of xdsl to be able to parse the remaining ops in the mlperf tiny benchmark.

Additionally, it adds some extra stuff to convert `arith.maxf` to `arith.maximumf` (similar for `arith.min`)

It looks like this finally supports the full set of benchmarks in linalg on memrefs.

Current TODOs
- [x] un-mess up the makefile indentation
- [x] add CI jobs
- [x] Update version of xdsl in requirements
- [x] Add CI for extra passes?


runtime/Makefile.rules
```diff
@@ -2,7 +2,7 @@ MAKEFILE_RULES_DIRNAME := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 
 # Courtesy of Federico Ficarelli
 
-# NOTE: we are using  absolute paths from the Docker container
+# NOTE: we are using	absolute paths from the Docker container
```
jorendumoulin, line 5: ?

jorendumoulin (approved):
Just change imumf

PR #66, opened 2024-01-24
replace deprecated ops

jorendumoulin:

PR #70, opened 2024-02-01
Replace deprecated memref constructor

JosseVanDelm:
It appears #66 did leave one deprected memref constructor :)


jorendumoulin (approved):
oops, thank you!!

PR #72, opened 2024-02-01
compiler: extend kernel matching

jorendumoulin:
This is the exact same PR as #41, but this one i will actually merge into main instead of some stale branch 🙃


JosseVanDelm (approved):
ok lol

PR #73, opened 2024-02-01
mlperf tiny: Lower to object file

JosseVanDelm:
This clears the pipeline to go directly to a RISC-V object file, yet networks are not actually run yet.
* A small error was found in the emission of `memref.get_global`s , where `name` and not `name_` was passed to the newly created op.
* There was a function call mismatch, where emitted function calls and declarations were incompatible with the shapes of various different calls. This was solved by declaring the functions with dynamic shapes and inserting casts to dynamic shapes.
* A pass (in the makefile) was integrated (with sed) to deal with minf, maxf to minimumf, maximumf conversion
* The `--clear-memory-space` pass was altered to only remove TSL layout attributes, other layout info is untouched.
* `--lower-affine` and `--expand-strided-metadata` passes were added to lower to LLVM dialect, otherwise unrealized_conversion_casts remain.
* A vanilla pipeline was added for each mlperf tiny benchmark that does not rely on snax-opt to compile (these object_files are also not run yet)

### DIFF
--- a/.github/workflows/build-run-mlperf-tiny.yml
+++ b/.github/workflows/build-run-mlperf-tiny.yml
@@ -28,20 +28,20 @@ jobs:
       - name: Anomaly detection
         run: |
           export PATH=/opt/python3.11/bin:$PATH
-          make ad01_int8.snax-opt.mlir
+          make ad01_int8.o
         working-directory: kernels/mlperf_tiny_ad01
       - name: Image classfication
         run: |
           export PATH=/opt/python3.11/bin:$PATH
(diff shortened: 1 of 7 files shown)


jorendumoulin (approved):

jorendumoulin (approved):

JosseVanDelm:
I'm remarking this as a draft, because there are some things that are still missing to do the full lowering.
We can use this to clear the pipeline somewhat, to branch off to other PR's

jorendumoulin (approved):
Very nice! Could you maybe just update the pr comment to reflect the solutions in the PR?

PR #75, opened 2024-02-07
split up cast realization pass

jorendumoulin:
This PR splits up the set-memory-layout pass into two:

1. `set-memory-layout`: insert casts
2. `realize-memory-space-casts`

This is because we want to do stuff inbetween the two passes as well. 
For example, based on the layer to which an accelerator is dispatched, we can decide to which memory space it will be cast. After this, we decide on a given memory layout. After this, we can realize the memory space casts and reshuffling operations at once.


JosseVanDelm (approved):
LGTM

PR #76, opened 2024-02-07
add memory layout pass

jorendumoulin:
This PR implements an initial pass for enforcing memory layouts when operations are disptached to certain accelerators. Currently, the implementation is very naive. It imposes a specific memory layout on the input and output of the linalg operation dispatched to snax_qgemm by inserting layout casting ops. In the future, the memory layout will be selected in a more intelligent, automatic way.

### DIFF
--- a/compiler/transforms/set_memory_layout.py
+++ b/compiler/transforms/set_memory_layout.py
@@ -8,7 +8,7 @@ from xdsl.pattern_rewriter import (
     op_type_rewrite_pattern,
 )
 
-from compiler.dialects.snax import ReShuffleOp
+from compiler.dialects.snax import LayoutCast
 from compiler.dialects.tsl import TiledStridedLayoutAttr
 from compiler.ir.tsl import Stride, TiledStride, TiledStridedLayout
 
@@ -62,15 +62,15 @@ class AddMemoryLayout(RewritePattern):
             )
 
             # insert reshuffling ops
-            new_input_a = ReShuffleOp.from_type_and_target_layout(
+            new_input_a = LayoutCast.from_type_and_target_layout(
                 linalg_op.inputs[0], tsl_input_a
             )
 
-            new_input_b = ReShuffleOp.from_type_and_target_layout(
+            new_input_b = LayoutCast.from_type_and_target_layout(
                 linalg_op.inputs[1], tsl_input_b
             )
 
-            new_output = ReShuffleOp.from_type_and_target_layout(
+            new_output = LayoutCast.from_type_and_target_layout(
                 linalg_op.outputs[0], tsl_output
             )
 
--- a/compiler/transforms/set_memory_layout.py
+++ b/compiler/transforms/set_memory_layout.py
@@ -26,7 +26,8 @@ class AddMemoryLayout(RewritePattern):
 
     @op_type_rewrite_pattern
     def match_and_rewrite(self, linalg_op: linalg.Generic, rewriter: PatternRewriter):
-        # check if operation is dispatched via library call
+        # check if operation is dispatched via library call, as set by e.g.
+        # the dispatch-kernels pass
         if linalg_op.library_call is None:
             return
         else:
--- a/compiler/transforms/set_memory_layout.py
+++ b/compiler/transforms/set_memory_layout.py
@@ -18,8 +18,10 @@ class AddMemoryLayout(RewritePattern):
     This class represents a rewrite pattern for adding memory layout to a
     linalg operation. The implementation is very naive. It imposes a specific
     memory layout on the input and output of the linalg operation dispatched
-    to snax_qgemm by inserting reshuffling ops. In the future, the memory
+    to snax_qgemm by inserting layout_cast ops. In the future, the memory
     layout will be selected in a more automatic way.
+
+    Note: currently, only snax_qgemm is supported.
     """
 
     @op_type_rewrite_pattern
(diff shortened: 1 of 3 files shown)


compiler/transforms/set_memory_layout.py
```diff
@@ -0,0 +1,99 @@
...
+from compiler.ir.tsl import Stride, TiledStride, TiledStridedLayout
+
+
+class AddMemoryLayout(RewritePattern):
+    """
+    This class represents a rewrite pattern for adding memory layout to a
+    linalg operation. The implementation is very naive. It imposes a specific
+    memory layout on the input and output of the linalg operation dispatched
+    to snax_qgemm by inserting reshuffling ops. In the future, the memory
+    layout will be selected in a more automatic way.
```
JosseVanDelm, line 22:
```suggestion
    layout will be selected in a more automatic way.
    
    Note: currently only snax_qgemm is supported.
```

compiler/transforms/set_memory_layout.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, linalg_op: linalg.Generic, rewriter: PatternRewriter):
+        # check if operation is dispatched via library call
```
JosseVanDelm, line 27:
```suggestion
        # check if operation is dispatched via library call, as set by e.g. the linalg-to-library-call pass
```

compiler/transforms/set_memory_layout.py
```diff
@@ -0,0 +1,99 @@
...
+            tsl_output = TiledStridedLayoutAttr(
+                TiledStridedLayout(
+                    [
+                        TiledStride([Stride(256, None), Stride(4, 8)]),
+                        TiledStride([Stride(None, None), Stride(32, 8)]),
+                    ]
+                )
+            )
+
+            # insert reshuffling ops
```
JosseVanDelm, line 64:
```suggestion
            # insert layout_cast ops
```

compiler/transforms/set_memory_layout.py
```diff
@@ -0,0 +1,99 @@
...
+from compiler.dialects.tsl import TiledStridedLayoutAttr
+from compiler.ir.tsl import Stride, TiledStride, TiledStridedLayout
+
+
+class AddMemoryLayout(RewritePattern):
+    """
+    This class represents a rewrite pattern for adding memory layout to a
+    linalg operation. The implementation is very naive. It imposes a specific
+    memory layout on the input and output of the linalg operation dispatched
+    to snax_qgemm by inserting reshuffling ops. In the future, the memory
```
JosseVanDelm, line 21:
```suggestion
    to snax_qgemm by inserting layout_cast ops. In the future, the memory
```

compiler/transforms/set_memory_layout.py
```diff
@@ -0,0 +1,99 @@
...
+            )
+
+            tsl_output = TiledStridedLayoutAttr(
+                TiledStridedLayout(
+                    [
+                        TiledStride([Stride(256, None), Stride(4, 8)]),
+                        TiledStride([Stride(None, None), Stride(32, 8)]),
+                    ]
+                )
+            )
```
JosseVanDelm, line 62: Would it make sense here to already write this in terms of parametrized banks?

tests/filecheck/transforms/set-memory-layout.mlir
```diff
@@ -0,0 +1,43 @@
...
+// CHECK-NEXT:     ^1(%arg3 : i8, %arg4 : i8, %arg5 : i32, %arg6 : i32, %arg7 : i32):
+// CHECK-NEXT:       %5 = "arith.extsi"(%arg3) : (i8) -> i32
+// CHECK-NEXT:       %6 = "arith.subi"(%5, %arg5) : (i32, i32) -> i32
+// CHECK-NEXT:       %7 = "arith.extsi"(%arg4) : (i8) -> i32
+// CHECK-NEXT:       %8 = "arith.subi"(%7, %arg6) : (i32, i32) -> i32
+// CHECK-NEXT:       %9 = "arith.muli"(%6, %8) : (i32, i32) -> i32
+// CHECK-NEXT:       %10 = "arith.addi"(%arg7, %9) : (i32, i32) -> i32
+// CHECK-NEXT:       "linalg.yield"(%10) : (i32) -> ()
+// CHECK-NEXT:     }) {"library_call" = "snax_qgemm"} : (memref<?x128xi8, #tsl.tsl<[?, 8] -> (?, 8), [?, 8] -> (256, 1)>, 1 : i32>, memref<128x128xi8, #tsl.tsl<[?, 8] -> (256, 1), [?, 8] -> (?, 8), offset: 64>, 1 : i32>, i32, i32, memref<?x128xi32, #tsl.tsl<[?, 8] -> (256, 4), [?, 8] -> (?, 32)>, 1 : i32>) -> ()
+// CHECK-NEXT:     "func.return"(%arg2) : (memref<?x128xi32, 1 : i32>) -> ()
```
JosseVanDelm, line 40: Shouldn't this return %4 instead? It feels like there's a tsl_layout -> no_layout missing here?

JosseVanDelm (changes requested):

compiler/transforms/set_memory_layout.py
jorendumoulin: The `linalg-to-library-call` does not set the library call? Maybe this is not a good name, but the pass transforms a `linalg` (with a library call) to an actual library call (insert `funcOps`)

compiler/transforms/set_memory_layout.py
jorendumoulin: I don't think this is worth it, I want to get rid of this code as soon as possible, + this will change a lot for experiments with different data layouts and stuff

JosseVanDelm (approved):
Thanks! I updated my comments!

PR #77, opened 2024-02-07
compiler: add realize-memref-casts pass

jorendumoulin:
This PR implements the `realize-memref-casts` pass.
This pass replaces the previous `realize-memory-space-casts` pass, and now includes both `memory_space_cast` and `snax_reshuffle` operations.
The cast transformations are handled by allocating new memrefs in the target layout space and inserting `memref.copy`s, assuming the transformation is handled by further lowering passes.

### DIFF
--- a/compiler/transforms/realize_memref_casts.py
+++ b/compiler/transforms/realize_memref_casts.py
@@ -1,0 +1,128 @@
+from xdsl.dialects import arith, builtin, func, linalg, memref
+from xdsl.dialects.memref import MemorySpaceCast
+from xdsl.ir import MLContext, Operation, OpResult
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import (
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+from compiler.dialects.snax import ReShuffleOp
+
+
+def is_cast_op(op: Operation) -> bool:
+    return isinstance(op, MemorySpaceCast) or isinstance(op, ReShuffleOp)
+
+
+class RealizeMemrefCasts(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(
+        self, op: MemorySpaceCast | ReShuffleOp, rewriter: PatternRewriter
+    ):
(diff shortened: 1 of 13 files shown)


compiler/transforms/realize_memref_casts.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # due to previous passes, it is common for multiple memref casting
+        # ops to be chained together. For now all the transformations are handled
+        # by the DMA which can access all memory spaces, and handle all transformations
+        # so we can fuse all the casting operations together.
+
+        # keep track of ops to add
+        ops_to_add = []
+
+        # if the source of the memref cast is another reshuffle op,
```
JosseVanDelm, line 37:
```suggestion
        # if the source of the memref cast is another layout_cast op,
```

JosseVanDelm: Sorry for forcing you to do all of these refactorings :(

compiler/transforms/realize_memref_casts.py
```diff
@@ -0,0 +1,128 @@
...
+)
+
+from compiler.dialects.snax import LayoutCast
+
+
+def is_cast_op(op: Operation) -> bool:
+    return isinstance(op, MemorySpaceCast) or isinstance(op, LayoutCast)
+
+
+class RealizeMemrefCasts(RewritePattern):
```
JosseVanDelm, line 19: docstring pls :pray:

compiler/transforms/realize_memref_casts.py
```diff
@@ -0,0 +1,128 @@
...
+        # create alloc op
+        alloc_op = memref.Alloc.get(
+            dest_type.get_element_type(),
+            64,  # default 64 alignment
+            dest_type.get_shape(),
+            dynamic_sizes=dyn_operands,
+            layout=dest_type.layout,
+            memory_space=dest_type.memory_space,
+        )
+        ops_to_add.append(alloc_op)
```
JosseVanDelm, line 76: Would it make sense to add the snax.alloc op here directly?

JosseVanDelm (approved):
Looks good!

PR #79, opened 2024-02-08
Extend TSL classes

jorendumoulin:
This PR extends the tsl with some useful functions necessary for better transformation support.
The algorithm to transform layouts only works from tsl -> tsl.
For memrefs with no layout or a simple stridedlayoutattr, it would therefore be useful to create tsl representations
of these layouts to still be able to use the current transformation algorithm


JosseVanDelm (approved):
Looks good!

PR #82, opened 2024-02-13
fix docker build

jorendumoulin:


jorendumoulin:
should be solved by kuleuven-micas/snitch_cluster#78

PR #83, opened 2024-02-15
Apply offset in transform copies

jorendumoulin:
This PR implements the offset in stridedlayouts and tiledstridedlayouts when copying with the transform copy. Before, they were simply ignored


JosseVanDelm (approved):
lgtm!

PR #84, opened 2024-02-15
add snax.alloc op

jorendumoulin:
This PR adds the `snax.alloc` operation

This also includes the util class for the `LLVMMemrefDescriptor`

### DIFF
--- a/compiler/dialects/snax.py
+++ b/compiler/dialects/snax.py
@@ -2,9 +2,19 @@ from __future__ import annotations
 
 from typing import cast
 
+from xdsl.dialects.builtin import IntegerType, NoneAttr
+from xdsl.dialects.llvm import LLVMPointerType
 from xdsl.dialects.memref import MemRefType, UnrankedMemrefType
 from xdsl.ir import Attribute, Dialect, Operation, SSAValue
-from xdsl.irdl import IRDLOperation, irdl_op_definition, operand_def, result_def
+from xdsl.irdl import (
+    IRDLOperation,
+    Operand,
+    OpResult,
+    irdl_op_definition,
+    operand_def,
+    opt_prop_def,
+    result_def,
+)
 from xdsl.utils.exceptions import VerifyException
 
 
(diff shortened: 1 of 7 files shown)


JosseVanDelm (commented):
Hi, nice PR, but this seems a bit like premature lowering to me, and I think it will make it difficult in later passes (a lot of unrealized_conversion casts will have to be inserted). 

Why are we immediately outputting an LLVM ptr?

To me it makes more sense to make an alternative to `memref.alloc` instead of an alternative to `llvm.call @malloc`

It seems a bit more idiomatic to just write our own `finalize-snax-alloc` pass, which lowers to the appropriate, `llvm.call` or even `func.call`? Then afterwards we can still (and have to) strip the TSLs before going llvm memref lowering passes anyway, right?

jorendumoulin:
Good point, thanks, I initially thought it would be necessary to directly output an llvm.ptr, but on second thought a memref is probably a better idea. I'll rethink this a bit!

jorendumoulin:
On third thought, I revert back to my original opinion that we should not output a memref. If we would output a memref type, a `memref.alloc` is sufficient in all cases. We also need something lower than a memref, because of the following problem:

Consider the following example:

```mlir
module {
  func.func @copy_to_alloc(%arg0: memref<64x64xi32>) {
    %alloc = memref.alloc() {alignment = 64 : i64} : memref<64x64xi32, #my_tsl, 1 : i32>
    memref.copy %arg0, %alloc : memref<64x64xi32> to memref<64x64xi32, #my_tsl, 1 : i32>
    return
  }
}
```

we can lower this to snax.alloc with a custom alloc size derived from the custom tsl layout:

```mlir
module {
  llvm.func @_mlir_memref_to_llvm_alloc(i32) -> !llvm.ptr
  func.func @copy_to_alloc(%arg0: memref<64x64xi32>) {
    %c50_i32 = arith.constant 50 : i32
    %0 = "snax.alloc"(%c50_i32) {alignment = 64 : i64} : (i32) -> memref<64x64xi32, #my_tsl, 1 : i32>
    memref.copy %arg0, %0 : memref<64x64xi32> to memref<64x64xi32, #my_tsl, 1 : i32>
    return
  }
}
```

however, lowering further to llvm runs into unrealized_conversion_cast from memref to the llvm.struct memref descriptor:

```mlir
%1 = "snax.alloc"(%c50_i32) {alignment = 64 : i64} : (i32) -> memref<64x64xi32, strided<[6, 1]>, 1 : i32>
%2 = builtin.unrealized_conversion_cast %1 : memref<64x64xi32, strided<[6, 1]>, 1 : i32> to !llvm.struct<(ptr<1>, ptr<1>, i32, array<2 x i32>, array<2 x i32>)>
```

so, we have to lower the snax alloc to produce the memref descriptor, but then, what is the use of the snax.alloc operation? The custom alloc size has been derived, but this is also still contained in the layout attribute of the memref.

For this reason, I don't think we should output a memref.
Either one of these seems better:
  a bare pointer, which we can use to construct an llvm.struct memref descriptor in a lowering pass
  an llvm.struct memref descriptor directly

For now, i prefer to output the llvm.struct memref descriptor directly. We then also insert an `unrealized_conversion_cast` from `struct->memref`. When the other operations get lowerd to llvm, the cast becomes obsolete and our problems are solved. Do you agree this is the way to go?

compiler/dialects/snax.py
```diff
@@ -64,4 +74,97 @@ def verify_(self) -> None:
...
+        if not isinstance(self.result.type.types, ArrayAttr):
+            raise VerifyException("Expected result type to have an ArrayAttr")
+
+        type_iter = iter(self.result.type.types.data)
+
+        if not isinstance(next(type_iter), LLVMPointerType):
+            raise VerifyException(
+                "Invalid Memref Descriptor: "
+                + "Expected first element to be LLVMPointerType"
+            )
```
JosseVanDelm, line 129: Can't you make a helper function that:
```python
def descriptor_check(type_instance , pos, type_check)        
     if not isinstance(next(type_iter), LLVMPointerType):
            raise VerifyException(
                "Invalid Memref Descriptor: "
                + f"Expected element at pos {pos} to be {type_check.__name__}"
            )
```
And then refactor the rest?

JosseVanDelm (changes requested):
In general this is good! but I think it can be improved a bit :smile: 
Can you add interop tests in your next PR - to make sure the unrealized_casts are indeed removed by `--reconcile-unrealized-casts`
Let me know what you think of my feedback!

jorendumoulin:
I refactored all the memref descriptor stuff to a brand new `LLVMMemrefDescriptor` class!

JosseVanDelm (approved):
:heart_eyes:

Issue #22, opened 2023-11-21
Add a default MLIR --> LLVM flow


JosseVanDelm:
There used to be a default MLIR --> LLVM flow before #12 , but it got overshadowed with the new flow.
This gives us no baseline to compare with

PR #85, opened 2024-02-15
Add vanilla MLIR flows

JosseVanDelm:
This PR adds flows that don't use any of the middle-end optimizations that we have made with xDSL.

You can now make the binaries, by calling:
```sh
make network_name.no-snax-opt.o
```
Solves #22


jorendumoulin (commented):
What is the use of this? We can't run this anyway on the snitch cluster without some of the xDSL passes

jorendumoulin (approved):

PR #86, opened 2024-02-20
compiler: add memref-to-snax op

jorendumoulin:


compiler/transforms/memref_to_snax.py
```diff
@@ -0,0 +1,126 @@
...
+
+        if not isinstance(element_type, builtin.IntegerType):
+            return
+
+        # get the memory space
+        memory_space = memref_type.memory_space
+
+        # if the memory space is not defined, conversion to snax is not possible
+        if isinstance(memory_space, NoneAttr):
+            return
```
JosseVanDelm, line 42: nit: Feels like if we put this check in the beginning, we could early-exit sooner in a lot of cases?

compiler/transforms/memref_to_snax.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # get the memref type
+        memref_type: memref.MemRefType = alloc_op.memref.type
+
+        # get the element type
+        element_type = memref_type.get_element_type()
+
+        if not isinstance(element_type, builtin.IntegerType):
+            return
```
JosseVanDelm, line 35: Why is this a requirement? f64 and i64 have the same size, and thus allocation requirement, no?

JosseVanDelm (commented):
This looks good, I just don't fully see what's happening here:

> in tsl.get_bound_ops, there are some memref.dim ops created based on the input memref operation (memref.alloc in this case). However, as the memref.alloc gets deleted, this is no longer a valid strategy. We therefore include the option to pass on a custom list of ops which hold the shape of the memref type, which will not be deleted.

Can you give an example, or can we discuss in person?

Currently it feels a bit like a hack to me that leaves the IR in between memref and llvm?

jorendumoulin:
It is not a hack, but quite an improvement of the previous situation.
An example:

```
%5 = memref.alloc : memref<16x16xi32, #my_tsl>
```

For replacing the function with a `snax.alloc`, we need the bound ops, previously, these required the use of `dim` ops:

```
%5 = memref.alloc : memref<16x16xi32, #my_tsl>
%6 = memref.dim(0) %5
%7 = memref.dim(1) %5
// further bound ops
```

The bound ops are then used to calculate the size for the `snax.alloc` operation. The `snax.alloc` will replace the `memref.alloc` operations, which is why the `memref.dim` operations can no longer be used. We can't reference the `memref.dim` operations to the `snax.alloc`, because we need the `dim` operations before the `snax.alloc`, and the `snax.alloc` also does not produce a `memref`.

The solution to the problem is like this:
If the shapes are static, insert `arith.constant ops`
If they are dynamic, they are present as arguments of the `memref.alloc`, and we can use those ops.

So instead of letting the `tsl.get_bound_ops` generate dim ops which will turn out to be invalid, we now generate the ops ourselves with this strategy and pass them to the function.

PR #88, opened 2024-02-21
snax.alloc: add alignment property

jorendumoulin:
This PR adds an alignment property to the `snax.alloc` op
This is important to us if we want to align a memref with a specific memory bank

### DIFF
--- a/compiler/dialects/snax.py
+++ b/compiler/dialects/snax.py
@@ -112,7 +112,7 @@ class Alloc(IRDLOperation):
         descriptor = LLVMMemrefDescriptor.from_rank_and_integer_type(rank, integer_type)
 
         if not alignment:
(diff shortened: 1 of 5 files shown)


JosseVanDelm (approved):
Good for me, just know that a i32 alignment property is invalid upstream for `memref.allocs`!
```
<source>:13:10: error: 'memref.alloc' op attribute 'alignment' failed to satisfy constraint: 64-bit signless integer attribute whose minimum value is 0
  %022 = "memref.alloc"() <{"alignment" = 64 : i32, "operandSegmentSizes" = array<i32: 0, 0>}> : () -> memref<16x16xi32>
         ^
```

jorendumoulin:
I didn't know that, thanks for pointing it out! I changed the tests to use `i64`

PR #91, opened 2024-02-23
add shape operands to snax alloc op

jorendumoulin:
For futher lowering the `snax.alloc` operation, we need the shape information to populate the llvm memref descriptor. The shape operands are already inserted in the ir, this PR just includes them as actual operands of the `snax.alloc` op.

### DIFF
--- a/compiler/transforms/memref_to_snax.py
+++ b/compiler/transforms/memref_to_snax.py
@@ -61,6 +61,8 @@ class AllocOpRewrite(RewritePattern):
                 ops_to_add.append(shape_op)
                 shape_ops.append(shape_op)
 
+        shape_ops_arg = [x for x in shape_ops]
+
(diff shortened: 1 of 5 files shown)


compiler/transforms/memref_to_snax.py
```diff
@@ -61,6 +61,8 @@ def match_and_rewrite(self, alloc_op: memref.Alloc, rewriter: PatternRewriter):
                 ops_to_add.append(shape_op)
                 shape_ops.append(shape_op)
 
+        shape_ops_arg = [x for x in shape_ops]
+
```
JosseVanDelm, line 65: This is not doing anything, right? We could just pass shape_ops immediately, no?

JosseVanDelm (approved):
Okay! Just a small nit?

PR #92, opened 2024-02-23
limit lowering to snax.alloc to memory space L1

jorendumoulin:
To fix the errors in mlperf tiny benchmarks from #90


JosseVanDelm (approved):
Nice!

PR #99, opened 2024-03-01
dialects: (acc) Add acc dialect and acc-cse pass to simplify setups

AntonLydike:
`acc` dialect: (snax) accelerator setup and dispatch code, used to represent setup, launching and waiting for accelerator calls.

`acc-cse` pass: elide redundant setup operations from IR.

### DIFF
--- a/compiler/dialects/acc.py
+++ b/compiler/dialects/acc.py
@@ -22,7 +22,6 @@ from xdsl.irdl import (
     result_def,
     var_operand_def,
 )
-from xdsl.utils.hints import isa
 
 
 @irdl_attr_definition
@@ -84,6 +83,7 @@ class SetupOp(IRDLOperation):
     If acc.setup is called without any parameters, the resulting state is the
     "empty" state, that represents a state without known values.
     """
+
     name = "acc.setup"
 
     values = var_operand_def(Attribute)  # TODO: make more precise?
--- a/compiler/dialects/acc.py
+++ b/compiler/dialects/acc.py
@@ -22,6 +22,7 @@ from xdsl.irdl import (
     result_def,
     var_operand_def,
 )
+from xdsl.utils.hints import isa
 
 
 @irdl_attr_definition
@@ -76,6 +77,13 @@ class AwaitOp(IRDLOperation):
 
 @irdl_op_definition
 class SetupOp(IRDLOperation):
+    """
+    acc.setup writes values to a specific accelerators configuration and returns
+    a value representing the currently known state of that accelerator's config.
+
+    If acc.setup is called without any parameters, the resulting state is the
+    "empty" state, that represents a state without known values.
+    """
     name = "acc.setup"
 
     values = var_operand_def(Attribute)  # TODO: make more precise?
@@ -108,17 +116,21 @@ class SetupOp(IRDLOperation):
     def __init__(
         self,
         vals: Sequence[SSAValue],
-        param_names: Sequence[str],
+        param_names: Sequence[str] | Sequence[StringAttr],
         accelerator: str | StringAttr,
         in_state: SSAValue | Operation | None = None,
     ):
         if not isinstance(accelerator, StringAttr):
             accelerator = StringAttr(accelerator)
 
+        param_names_tuple: tuple[StringAttr, ...] = tuple(
+            StringAttr(name) if isinstance(name, str) else name for name in param_names
+        )
+
         super().__init__(
             operands=[vals, in_state],
             properties={
-                "param_names": ArrayAttr([StringAttr(x) for x in param_names]),
+                "param_names": ArrayAttr(param_names_tuple),
                 "accelerator": accelerator,
             },
             result_types=[StateType((accelerator,))],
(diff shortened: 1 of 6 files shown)


compiler/dialects/acc.py
```diff
@@ -0,0 +1,163 @@
...
+)
+
+
+@irdl_attr_definition
+class TokenType(ParametrizedAttribute, TypeAttribute):
+    """
+    Async token type for launched accelerator requests.
+    """
+
+    name = "acc.token"
```
jorendumoulin, line 33: The `TokenType` is empty: let me just check if I get the working of this correctly: The sole purpose of a token is to easily know which `acc.wait` and `acc.launch` are linked together, which is useful in some passes? This won't actually be something that is lowered, as the `acc.await` will be based on the accelerator id, not the token.

compiler/dialects/acc.py
```diff
@@ -0,0 +1,163 @@
...
+    def __init__(
+        self,
+        vals: Sequence[SSAValue],
+        param_names: Sequence[str],
+        accelerator: str | StringAttr,
+        in_state: SSAValue | Operation | None = None,
+    ):
+        if not isinstance(accelerator, StringAttr):
+            accelerator = StringAttr(accelerator)
+
```
jorendumoulin, line 117: Adding support for existing StringAttr params can be nice
```suggestion
        param_names: Sequence[str | StringAttr],
        accelerator: str | StringAttr,
        in_state: SSAValue | Operation | None = None,
    ):
        if not isinstance(accelerator, StringAttr):
            accelerator = StringAttr(accelerator)
        
        param_names = [
            StringAttr(param_name) if not isinstance(param_name, StringAttr) 
            else param_name 
            for param_name in param_names
        ]
```

compiler/transforms/acc_cse.py
```diff
@@ -0,0 +1,78 @@
...
+            (name, val) for name, val in op.iter_params() if prev_state.get(name) != val
+        ]
+
+        # Step 3: If no new params remain, yeet the whole op
+        if not new_params:
+            op.out_state.replace_by(op.in_state)
+            rewriter.erase_matched_op()
+            return
+
+        # Step 4: If all parameters change, we are a no-op
```
jorendumoulin, line 53:
```suggestion
        # Step 4: If all parameters change, do nothing
```

compiler/transforms/acc_cse.py
```diff
@@ -0,0 +1,78 @@
...
+    def match_and_rewrite(self, op: acc.SetupOp, rewriter: PatternRewriter, /):
+        # Step 1: Figure out previous state
+        prev_state = get_state_before(op)
+
+        # Step 2: Yeet out redundant stuff
+        new_params: list[tuple[str, SSAValue]] = [
+            (name, val) for name, val in op.iter_params() if prev_state.get(name) != val
+        ]
+
+        # Step 3: If no new params remain, yeet the whole op
```
jorendumoulin, line 47:
```suggestion
        # Step 3: If no parameters change, the whole op can be erased
```

compiler/transforms/acc_cse.py
```diff
@@ -0,0 +1,78 @@
...
+    return state
+
+
+class YeetSetupStuff(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: acc.SetupOp, rewriter: PatternRewriter, /):
+        # Step 1: Figure out previous state
+        prev_state = get_state_before(op)
+
+        # Step 2: Yeet out redundant stuff
```
jorendumoulin, line 42:
```suggestion
        # Step 2: Yeet out redundant stuff (= params that get set to the same value)
```

compiler/transforms/acc_cse.py
```diff
@@ -0,0 +1,78 @@
...
+
+        for name, val in parent.iter_params():
+            if name not in state:
+                state[name] = val
+
+        op = parent
+    return state
+
+
+class YeetSetupStuff(RewritePattern):
```
jorendumoulin, line 36: I love the name, but perhaps `AccSetupCse`, is a tiny bit more clear?

jorendumoulin (approved):
Looks good! I have some minor suggestions

compiler/dialects/acc.py
AntonLydike: Yes, pretty much. The token is used so we can trace which acccelerator to wait on. E.g. the launch launches an accelerator, and this way the await will always match the launch accelerator.

This also allows us to maybe infer the kernel runtimes, so we can push setup calls up before the previous await, saving more cycles!

compiler/dialects/acc.py
AntonLydike: Yeah, we had this before, but pyright can't infer the types. We can add it, but it's not needed so far, and the way we ended up doing it in xDSL was:

```py
param_names : Sequence[str] | Sequence[StringAttr]  # cannot be a mixed sequence

# ...
params = tuple(param_names) # `isa` doesn't support Sequence, so cast to tuple
if not isa(params, tuple[StringAttr, ...]):
    params = tuple(StringAttr(x) for x in params
```
But this is not really nice looking

PR #102, opened 2024-03-04
Lower linalg to `acc2`

AntonLydike:
WIP of the lowering from linalg to acc2 dialect


AntonLydike:
Couldn't fix the history mess, opened new branch

PR #103, opened 2024-03-04
Lower linalg to acc2 dialect

AntonLydike:
This PR contains:

 - Rename `acc` to `acc2`, as `acc` is already used for the openACC dialect in MLIR
 - A new `convert-linalg-to-acc` pass:
    - That converts linalg annotated with `library_call = "snax_hwpe_mult"` to acc2 dialect ops
    - That then connects the `!acc2.state` ssa values through the IR such that they conform to the control flow (currently handles `scf.if` but not `scf.for`)
 - Improvements to the `acc-cse` pass that is now able to infer states through `scf.if` ops. 
 - A test that depends on `mlir-opt` passes, we marked it `XFAIL` for now until we can update to a commit of xDSL that contains xdslproject/xdsl#2293

### DIFF
--- a/compiler/dialects/acc.py
+++ b/compiler/dialects/acc.py
@@ -1,14 +1,29 @@
 from __future__ import annotations
 
-from collections.abc import Iterable, Sequence
+from collections.abc import Iterable
 
 from xdsl.dialects.builtin import ArrayAttr, StringAttr
-from xdsl.ir import (Attribute, Dialect, Operation, ParametrizedAttribute,
-                     SSAValue, TypeAttribute)
-from xdsl.irdl import (AttrSizedOperandSegments, IRDLOperation, ParameterDef,
-                       VerifyException, irdl_attr_definition,
-                       irdl_op_definition, operand_def, opt_operand_def,
-                       prop_def, result_def, var_operand_def)
+from xdsl.ir import (
+    Attribute,
+    Dialect,
+    Operation,
+    ParametrizedAttribute,
+    SSAValue,
+    TypeAttribute,
+)
+from xdsl.irdl import (
+    AttrSizedOperandSegments,
+    IRDLOperation,
+    ParameterDef,
+    VerifyException,
+    irdl_attr_definition,
+    irdl_op_definition,
+    operand_def,
+    opt_operand_def,
+    prop_def,
+    result_def,
+    var_operand_def,
+)
 
 
 @irdl_attr_definition
@@ -133,8 +148,8 @@ class SetupOp(IRDLOperation):
 
     def __init__(
         self,
-        vals: Sequence[SSAValue | Operation],
-        param_names: Sequence[str] | Sequence[StringAttr],
+        vals: Iterable[SSAValue | Operation],
+        param_names: Iterable[str] | Iterable[StringAttr],
         accelerator: str | StringAttr,
         in_state: SSAValue | Operation | None = None,
     ):
--- a/compiler/dialects/acc.py
+++ b/compiler/dialects/acc.py
@@ -1,6 +1,11 @@
-from collections.abc import Iterable, Sequence
+from __future__ import annotations
 
-from xdsl.dialects.builtin import ArrayAttr, StringAttr
+from collections.abc import Iterable, Sequence
+from xdsl.parser import Parser
+from xdsl.printer import Printer
+
+from xdsl.traits import SymbolOpInterface
+from xdsl.dialects.builtin import ArrayAttr, StringAttr, SymbolRefAttr, DictionaryAttr, IntegerAttr, i32
 from xdsl.ir import (
     Attribute,
     Dialect,
@@ -30,7 +35,7 @@ class TokenType(ParametrizedAttribute, TypeAttribute):
     Async token type for launched accelerator requests.
     """
 
-    name = "acc.token"
+    name = "acc2.token"
 
 
 @irdl_attr_definition
@@ -39,7 +44,7 @@ class StateType(ParametrizedAttribute, TypeAttribute):
     Used to trace an accelerators CSR state through def-use chain
     """
 
-    name = "acc.state"
+    name = "acc2.state"
 
     accelerator: ParameterDef[StringAttr]
 
@@ -52,13 +57,23 @@ class LaunchOp(IRDLOperation):
     interfering with the Accelerator.
     """
 
-    name = "acc.launch"
+    name = "acc2.launch"
 
     state = operand_def(StateType)
 
     accelerator = prop_def(StringAttr)
 
     token = result_def()
+
+    def __init__(self, state: SSAValue | Operation):
+        state_val: SSAValue = SSAValue.get(state)
+        if not isinstance(state_val.type, StateType):
+            raise ValueError("`state` SSA Value must be of type `acc2.state`!")
+        super().__init__(
+            operands=[state],
+            properties={"accelerator": state_val.type.accelerator},
+            result_types=[TokenType()]
+        )
 
     def verify_(self) -> None:
         # that the state and my accelerator match
@@ -82,22 +97,25 @@ class AwaitOp(IRDLOperation):
     Blocks until the launched operation finishes.
     """
 
-    name = "acc.await"
+    name = "acc2.await"
 
     token = operand_def(TokenType)
+
+    def __init__(self, token: SSAValue | Operation):
+        super().__init__(operands=[token])
 
 
 @irdl_op_definition
 class SetupOp(IRDLOperation):
     """
-    acc.setup writes values to a specific accelerators configuration and returns
+    acc2.setup writes values to a specific accelerators configuration and returns
     a value representing the currently known state of that accelerator's config.
 
-    If acc.setup is called without any parameters, the resulting state is the
+    If acc2.setup is called without any parameters, the resulting state is the
     "empty" state, that represents a state without known values.
     """
 
-    name = "acc.setup"
+    name = "acc2.setup"
 
     values = var_operand_def(Attribute)  # TODO: make more precise?
     """
@@ -106,7 +124,7 @@ class SetupOp(IRDLOperation):
 
     in_state = opt_operand_def(StateType)
     """
-    The state produced by a previous acc.setup
+    The state produced by a previous acc2.setup
     """
 
     out_state = result_def(StateType)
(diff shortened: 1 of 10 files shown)


compiler/inference/trace_acc_state.py
```diff
@@ -0,0 +1,69 @@
...
+    states = []
+    for region in op.regions:
+        # we know the last op must be the yield
+        yield_op = region.block.last_op
+        assert isinstance(yield_op, scf.Yield)
+        # we know the yield op has the same number of operands as the
+        # scf.if has results, so [idx] must be defined
+        states.append(infer_state_of(yield_op.operands[idx]))
+    assert len(states) == 2
+    return tuple(states)
```
AntonLydike, line 65: Important change

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+                        state[res.type.accelerator.data] = res
+
+                # calling another function invalidates all states
+                elif isinstance(op, func.Call):
+                    state.clear()
+                elif isinstance(op, func.Return | scf.Yield):
+                    continue
+                else:
+                    raise RuntimeError(f"What is a {op}?")
+    return state
```
AntonLydike, line 182: Important change

compiler/inference/trace_acc_state.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def state_union(a: State, b: State) -> State:
+    return {k: a[k] for k in a if a[k] == b[k]}
```
jorendumoulin, line 69: I think this throws a `KeyError` if `k not in b`

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+            setup := acc.SetupOp(
+                [val for _, val in args], accelerator.fields, accelerator.name
+            )
+        )
+
+        # launch
+        rewriter.insert_op_before_matched_op(token := acc.LaunchOp(setup))
+
+        # await
+        rewriter.replace_matched_op(acc.AwaitOp(token))
```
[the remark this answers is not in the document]
AntonLydike: This pattern replaces linalg that is dispatched to `snax_hwpe_mult` with the `acc` dialect.

compiler/transforms/convert_linalg_to_acc.py
AntonLydike: This pattern connects all the generated `acc2` calls via def-use chain such that we can optimize them later on

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+class ConvertLinalgToAcceleratorPattern(RewritePattern):
+    module: builtin.ModuleOp
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: linalg.Generic, rewriter: PatternRewriter, /):
+        if op.library_call is None:
+            return
+
+        if op.library_call.data != "snax_hwpe_mult":
+            return
```
jorendumoulin, line 58: redundant check

compiler/inference/trace_acc_state.py
AntonLydike: This is a better version of the previously existent inference code that analyzes variables set in the state. This can now handle if/else blocks.

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+from compiler.dialects import acc
+
+
+class AcceleratorConfig:
```
jorendumoulin, line 17: This feels like it does not really belong here? `AcceleratorConfig` should also be an abstract class I feel, now it is confusing that `snax_hwpe_mult` is the name of the class

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+        rewriter.insert_op_before_matched_op(token := acc.LaunchOp(setup))
+
+        # await
+        rewriter.replace_matched_op(acc.AwaitOp(token))
+
+
+@dataclass
+class ConnectStatesThroughControlFlowPattern(RewritePattern):
+    walked_funcs: set[str] = field(default_factory=set)
+
```
JosseVanDelm, line 87: Add your comment as a docstring

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+        regions = container.regions
+    else:
+        regions = [container]
+
+    for region in regions:
+        for block in region.blocks:
+            for op in block.ops:
+                # arith, memref, linalg and test ops are not relevant to
+                # accelerator setup
+                if op.name.split(".")[0] in ("arith", "memref", "linalg", "test"):
```
jorendumoulin, line 109: hmmmm

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+                    continue
+                # handle acc2 dialect ops:
+                elif op.name.startswith("acc2."):
```
jorendumoulin, line 112: hmmmmmmmmmm

JosseVanDelm (changes requested):
Very nice work, but I think we can still clean this up a bit :smile:

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+                    # the setup op is the only relevant one for now:
+                    if isinstance(op, acc.SetupOp):
```
jorendumoulin, line 114: then just do this instead of `elif op.op.name.startswith("acc2.")`?

compiler/transforms/convert_linalg_to_acc.py
jorendumoulin: you can just leave this out, right?

jorendumoulin (changes requested):
Some really cool stuff!

I do have some small suggestions

compiler/dialects/acc.py
```diff
@@ -1,4 +1,6 @@
-from collections.abc import Iterable, Sequence
+from __future__ import annotations
```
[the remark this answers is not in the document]
AntonLydike: It allows us to reference a type while it is being constructed (e.g. have a typing annotation that says that a classmethod returns an instance of itself)

compiler/inference/trace_acc_state.py
```diff
@@ -0,0 +1,69 @@
...
+            return {name: val for name, val in setup_op.iter_params()}
+        case acc.SetupOp(in_state=st) as setup_op if st is not None:
+            in_state = infer_state_of(st)
+            in_state.update(dict(setup_op.iter_params()))
+            return in_state
+        case scf.If() as if_op:
+            return state_union(*infer_states_for_if(if_op, state_var))
+        case Block():
+            # TODO: maybe implement something here?
+            return {}
```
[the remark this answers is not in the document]
AntonLydike: If we want to add some information that e.g. passes state information across function borders, or a more generalized thing, idk.

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+                        assert isinstance(res.type, acc.StateType)
+                        state[res.type.accelerator.data] = res
+
+                # calling another function invalidates all states
+                elif isinstance(op, func.Call):
+                    state.clear()
+                elif isinstance(op, func.Return | scf.Yield):
+                    continue
+                else:
+                    raise RuntimeError(f"What is a {op}?")
```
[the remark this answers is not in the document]
AntonLydike: Possibly

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+
+        ptrs = [
+            (
+                [ptr := memref.ExtractAlignedPointerAsIndexOp.get(ref)],
+                ptr.aligned_pointer,
+            )
+            for ref in (a, b, c)
+        ]
+
+        return ptrs + [size]
```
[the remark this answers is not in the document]
AntonLydike: Yes, that is the plan eventually.

compiler/inference/trace_acc_state.py
AntonLydike: Yes, you are right!

compiler/transforms/convert_linalg_to_acc.py
AntonLydike: Yes, this was an abstract class once upon a time, but then we wanted to be a bit more concrete. Never made it more general again.

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+    container: Region | Operation, state: dict[str, SSAValue], rewriter: PatternRewriter
+) -> dict[str, SSAValue]:
+    if isinstance(container, Operation):
+        regions = container.regions
+    else:
+        regions = [container]
+
+    for region in regions:
+        for block in region.blocks:
+            for op in block.ops:
```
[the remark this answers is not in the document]
AntonLydike: No, we don't want to descend into regions of sub-ops.

compiler/transforms/convert_linalg_to_acc.py
AntonLydike: Yeah, I guess this should be cleaned up a bit further. I was somewhat unhinged when I wrote this code yesterday

compiler/transforms/convert_linalg_to_acc.py
```diff
@@ -0,0 +1,237 @@
...
+                                op.param_names,
+                                op.accelerator,
+                                state[accel],
+                            )
+                            rewriter.replace_op(op, new_op)
+                            op = new_op
+                        state[accel] = op.out_state
+                    else:
+                        pass
+                elif isinstance(op, scf.If):
```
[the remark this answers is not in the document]
jorendumoulin: I support Anton

compiler/transforms/convert_linalg_to_acc.py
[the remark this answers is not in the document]
JosseVanDelm: yes, okay to do in a next PR!

compiler/transforms/convert_linalg_to_acc.py
AntonLydike: I made it a tiny bit more palatable hopefully.

jorendumoulin (approved):
Thank you!

JosseVanDelm (approved):
:heart:

PR #104, opened 2024-03-05
Improve acc dialect optimizations

AntonLydike:
Stacked on top of #103 

This PR:

- Renames `acc-cse` to `acc-dedup` to better reflect what it's doing
- Adds a rewrite to hoist setup calls into `scf.if` operations

### DIFF
--- a/compiler/tools/snax_opt_main.py
+++ b/compiler/tools/snax_opt_main.py
@@ -7,7 +7,7 @@ from xdsl.xdsl_opt_main import xDSLOptMain
 from compiler.dialects.acc import ACC
 from compiler.dialects.snax import Snax
 from compiler.dialects.tsl import TSL
-from compiler.transforms.acc_cse import AccCse
+from compiler.transforms.acc_cse import AccDeduplicate
 from compiler.transforms.clear_memory_space import ClearMemorySpace
 from compiler.transforms.convert_linalg_to_acc import ConvertLinalgToAccPass
 from compiler.transforms.dispatch_kernels import DispatchKernels
@@ -55,7 +55,7 @@ class SNAXOptMain(xDSLOptMain):
             RealizeMemrefCastsPass.name, lambda: RealizeMemrefCastsPass
         )
         super().register_pass(MemrefToSNAX.name, lambda: MemrefToSNAX)
-        super().register_pass(AccCse.name, lambda: AccCse)
+        super().register_pass(AccDeduplicate.name, lambda: AccDeduplicate)
         super().register_pass(
             ConvertLinalgToAccPass.name, lambda: ConvertLinalgToAccPass
         )
--- a/compiler/tools/snax_opt_main.py
+++ b/compiler/tools/snax_opt_main.py
@@ -7,7 +7,7 @@ from xdsl.xdsl_opt_main import xDSLOptMain
 from compiler.dialects.acc import ACC
 from compiler.dialects.snax import Snax
 from compiler.dialects.tsl import TSL
-from compiler.transforms.acc_cse import AccDeduplicate
+from compiler.transforms.acc_dedup import AccDeduplicate
 from compiler.transforms.clear_memory_space import ClearMemorySpace
 from compiler.transforms.convert_linalg_to_acc import ConvertLinalgToAccPass
 from compiler.transforms.dispatch_kernels import DispatchKernels
--- a/compiler/transforms/acc_cse.py
+++ b/compiler/transforms/acc_cse.py
@@ -1,12 +1,8 @@
 from xdsl.dialects import builtin
 from xdsl.ir import MLContext, SSAValue
 from xdsl.passes import ModulePass
-from xdsl.pattern_rewriter import (
-    PatternRewriter,
-    PatternRewriteWalker,
-    RewritePattern,
-    op_type_rewrite_pattern,
-)
+from xdsl.pattern_rewriter import (PatternRewriter, PatternRewriteWalker,
+                                   RewritePattern, op_type_rewrite_pattern)
 
 from compiler.dialects import acc
 from compiler.inference.trace_acc_state import infer_state_of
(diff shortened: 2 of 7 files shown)


compiler/transforms/acc_cse.py
```diff
@@ -53,15 +54,90 @@ def match_and_rewrite(self, op: acc.SetupOp, rewriter: PatternRewriter, /):
         )
 
 
-class AccCse(ModulePass):
+class HoistSetupCallsIntoConditionals(RewritePattern):
     """
-    Common subexpression elimination for the `acc` (accelerator) dialect.
+    Hoists setup calls into scf.if blocks.
```
jorendumoulin, line 59:
```suggestion
    Hoists setup calls into scf.if blocks, preceding the setup calls.
```
(it took a while before I realized this)

compiler/transforms/acc_cse.py
```diff
@@ -53,15 +54,90 @@ def match_and_rewrite(self, op: acc.SetupOp, rewriter: PatternRewriter, /):
...
 
-    This pass rewrites acc dialect operations to find and simplify redundant
-    setup calls.
+    This will never result in additional setup calls and only
+    reduced the number of fields that are set up.
+    (proof left as an exercise to the reader)
+
+    Things we need to worry about:
+
+    - not inserting a setup op before the originally produced state
```
jorendumoulin, line 67:
```suggestion
    - we can not insert a setup op before the originally produced state
```

compiler/transforms/acc_cse.py
```diff
@@ -53,15 +54,90 @@ def match_and_rewrite(self, op: acc.SetupOp, rewriter: PatternRewriter, /):
...
+        assert isinstance(old_in_state, OpResult)
+
+        # Step 1: Check that it's legal to move:
+        uses = tuple(
+            use for use in op.in_state.uses if isinstance(use.operation, acc.LaunchOp)
+        )
+        # if we have some launch ops, we need to investigate further:
+        if len(uses) > 0:
+            # if there is a launch outside the scf.if
+            launch_op = uses[0].operation
```
jorendumoulin, line 87: what about the other uses? like `uses[1]`

compiler/transforms/acc_cse.py
```diff
@@ -53,15 +54,90 @@ def match_and_rewrite(self, op: acc.SetupOp, rewriter: PatternRewriter, /):
...
+            # check that it is in the same block (if not bail out)
+            # TODO: better check this case?
+            if launch_op.parent_block() is not op.parent_block():
+                return
```
jorendumoulin, line 91:
```mlir
scf.if %b  {
  acc.launch
} else {
  ...
}
acc.setup
```

Can you not still lift the setup op in the if region? I think you can make an exception if the `launch_op.parent_block()` is part of the `scf.if` under consideration

jorendumoulin (changes requested):
Can you also add some filechecks? That would make this much more clear

compiler/transforms/acc_cse.py
AntonLydike: I think you missed one tiny detail, the IR would look like this:

```mlir
%state_before_if = acc.setup()
acc.launnch(%state_before_if)
%state_after_if = scf.if (%i1) -> !acc.state {
  %s1 = acc.setup(%state_before_if)
  acc.launch(%s1)
  scf.yield %s1
} else {
  scf.yield %state_before_if
}

%s_new = acc.setup(%state_after_if) // <- this is our `op`
acc.launch(%s_new)
```

Here we are looking at the `%s_new = acc.setup(%state_after_if)` as our `op` variable.

The input state is the variable `%state_after_if`, and we don't find the `acc.launch(%s1)` op as a use of it.

The only way a launch op use of the `%state_after_if` cannot share a block with our op is if it is inside another nested op *after* the `if`:

```mlir
%state_before_if = acc.setup()
acc.launnch(%state_before_if)
%state_after_if = scf.if (%i1) -> !acc.state {
// ...

// either here (illegal):
some.op() {
  acc.launch(%state_after_if)
}

%s_new = acc.setup(%state_after_if) // <- this is our `op`
acc.launch(%s_new)



// or here (legal):
some.op() {
  acc.launch(%state_after_if)
}
```

But accounting for this structure is not in scope for this PR

jorendumoulin: ah, yes, this makes a lot of sense. thank you for the elaboration!

jorendumoulin (approved):
Thank you, this looks good!

compiler/tools/snax_opt_main.py
```diff
@@ -7,7 +7,7 @@
 from compiler.dialects.acc import ACC
 from compiler.dialects.snax import Snax
 from compiler.dialects.tsl import TSL
-from compiler.transforms.acc_cse import AccCse
+from compiler.transforms.acc_cse import AccDeduplicate
```
JosseVanDelm, line 10:
```suggestion
from compiler.transforms.acc_dedup import AccDeduplicate
```

compiler/transforms/acc_cse.py
```diff
@@ -53,15 +54,96 @@ def match_and_rewrite(self, op: acc.SetupOp, rewriter: PatternRewriter, /):
...
-    name = "acc-cse"
+    name = "acc-dedup"
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
-        PatternRewriteWalker(SimplifyRedundantSetupCalls()).rewrite_module(op)
+        PatternRewriteWalker(
+            GreedyRewritePatternApplier(
+                [
+                    HoistSetupCallsIntoConditionals(),
+                    SimplifyRedundantSetupCalls(),
```
JosseVanDelm, line 145: Can we not use these passes separately?

JosseVanDelm (approved):
Small nit,
Also, should we make these available as separate passes?
Seems easier to measure impact of the optimization later?

compiler/transforms/acc_cse.py
AntonLydike: Yes

compiler/tools/snax_opt_main.py
AntonLydike: Yes

PR #106, opened 2024-03-06
Add backend for acc2 dialect

AntonLydike:
This PR introduces a new pass `convert-acc-to-csr` that converts `acc` dialect ops to llvm inline assembly setting CSR register values.

It also adds a `acc2.accelerator` operation that defines acclerator specific addresses and such, which is used by the lowering.

### DIFF
--- a/compiler/dialects/acc.py
+++ b/compiler/dialects/acc.py
@@ -236,23 +236,16 @@ class AcceleratorOp(IRDLOperation):
 
     launch_addr = prop_def(IntegerAttr)
 
-    barrier_enable = prop_def(
+    barrier_sw_barrier = prop_def(
         IntegerAttr
     )  # TODO: this will be reworked in a later version
-
-    barrier_trigger = prop_def(
-        IntegerAttr
-    )  # TODO: this will be reworked in a later version
-
-    # TODO: handle the await/poll stuff
 
     def __init__(
         self,
         name: str | StringAttr | SymbolRefAttr,
         fields: dict[str, int] | DictionaryAttr,
         launch: int | IntegerAttr,
-        barrier_enable: int | IntegerAttr,
-        barrier_trigger: int | IntegerAttr,
+        barrier_sw_barrier: int | IntegerAttr,
     ):
         if not isinstance(fields, DictionaryAttr):
             fields = DictionaryAttr(
@@ -270,15 +263,10 @@ class AcceleratorOp(IRDLOperation):
                     if not isinstance(launch, IntegerAttr)
                     else launch
                 ),
-                "barrier_enable": (
-                    IntegerAttr(barrier_enable, i32)
-                    if not isinstance(barrier_enable, IntegerAttr)
-                    else barrier_enable
-                ),
-                "barrier_trigger": (
-                    IntegerAttr(barrier_trigger, i32)
-                    if not isinstance(barrier_trigger, IntegerAttr)
-                    else barrier_trigger
+                "barrier_sw_barrier": (
+                    IntegerAttr(barrier_sw_barrier, i32)
+                    if not isinstance(barrier_sw_barrier, IntegerAttr)
+                    else barrier_sw_barrier
                 ),
             }
         )
--- a/compiler/dialects/acc.py
+++ b/compiler/dialects/acc.py
@@ -236,16 +236,14 @@ class AcceleratorOp(IRDLOperation):
 
     launch_addr = prop_def(IntegerAttr)
 
-    barrier_sw_barrier = prop_def(
-        IntegerAttr
-    )  # TODO: this will be reworked in a later version
+    barrier = prop_def(IntegerAttr)  # TODO: this will be reworked in a later version
 
     def __init__(
         self,
         name: str | StringAttr | SymbolRefAttr,
         fields: dict[str, int] | DictionaryAttr,
         launch: int | IntegerAttr,
-        barrier_sw_barrier: int | IntegerAttr,
+        barrier: int | IntegerAttr,
     ):
         if not isinstance(fields, DictionaryAttr):
             fields = DictionaryAttr(
@@ -263,10 +261,10 @@ class AcceleratorOp(IRDLOperation):
                     if not isinstance(launch, IntegerAttr)
                     else launch
                 ),
-                "barrier_sw_barrier": (
-                    IntegerAttr(barrier_sw_barrier, i32)
-                    if not isinstance(barrier_sw_barrier, IntegerAttr)
-                    else barrier_sw_barrier
+                "barrier": (
+                    IntegerAttr(barrier, i32)
+                    if not isinstance(barrier, IntegerAttr)
+                    else barrier
                 ),
             }
         )
--- a/compiler/dialects/acc.py
+++ b/compiler/dialects/acc.py
@@ -1,6 +1,13 @@
 from collections.abc import Iterable
 
-from xdsl.dialects.builtin import ArrayAttr, StringAttr
+from xdsl.dialects.builtin import (
+    ArrayAttr,
+    DictionaryAttr,
+    IntegerAttr,
+    StringAttr,
+    SymbolRefAttr,
+    i32,
+)
 from xdsl.ir import (
     Attribute,
     Dialect,
@@ -22,6 +29,9 @@ from xdsl.irdl import (
     result_def,
     var_operand_def,
 )
+from xdsl.parser import Parser
+from xdsl.printer import Printer
+from xdsl.traits import SymbolOpInterface
 
 
 @irdl_attr_definition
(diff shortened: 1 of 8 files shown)


AntonLydike:
More comments in code coming to you tomorrow, pinky promise!

compiler/dialects/acc.py
```diff
@@ -189,12 +212,98 @@ def verify_(self) -> None:
...
 
 
+class AcceleratorSymbolOpTrait(SymbolOpInterface):
+    def get_sym_attr_name(self, op: Operation) -> StringAttr | None:
+        assert isinstance(op, AcceleratorOp)
+        return StringAttr(op.name_prop.string_value())
+
+
+@irdl_op_definition
+class AcceleratorOp(IRDLOperation):
```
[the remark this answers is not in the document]
jorendumoulin: The accelerator ops would then not be inserted by the compiler based on the operations in the file, but rather in a (separate) target_configuration, which includes a trait for every accelerator in the system.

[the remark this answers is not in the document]
AntonLydike: Not sure what you mean by global trait. Do you mean an attribute alias? (similar to the way you handle the strides for the affine maps?)

I guess it could be done that way too, having a (verified) unique operation just encodes better (imo) that a given accelerator always has the same fields. If you attach this info via attributes you enable mistakes like different attributes referring to different fields.

compiler/transforms/convert_acc_to_csr.py
```diff
@@ -0,0 +1,260 @@
...
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+from xdsl.traits import SymbolTable
+
+from compiler.dialects import acc
+
+
+@dataclass
+class LowerAccPattern(RewritePattern, ABC):
```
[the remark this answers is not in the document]
AntonLydike: How about `LowerAccBasePattern`, as it's the base class for the acc2 dialecct lowering patterns

JosseVanDelm (commented):
Hi sorry I started reviewing and stopped halfway hahaha.
I'm going to add some more detailed comments later.

compiler/transforms/convert_acc_to_csr.py
```diff
@@ -0,0 +1,254 @@
...
+                barrier_sw_barrier := arith.Constant(acc_op.barrier_sw_barrier),
+                zero := arith.Constant(builtin.IntegerAttr.from_int_and_width(0, 5)),
+                # FIXME: How to clobber a0?
+                llvm.InlineAsmOp(
+                    (
+                        "\n"
+                        "  csrr a0, $0\n"
+                        "1:\n"
+                        "  bnez a0, 1b\n"
+                        "  csrwi 0x3c5, $1\n"  # Weird clear routine
```
AntonLydike, line 134: We now have a hardcoded CSR address in here, is this what we want?

compiler/dialects/acc.py
```diff
@@ -189,12 +212,86 @@ def verify_(self) -> None:
...
+
+    name_prop = prop_def(SymbolRefAttr, prop_name="name")
+
+    fields = prop_def(DictionaryAttr)
+
+    launch_addr = prop_def(IntegerAttr)
+
+    barrier_sw_barrier = prop_def(
+        IntegerAttr
+    )  # TODO: this will be reworked in a later version
```
AntonLydike, line 241: This attribute now has a duplicate "barrier" in the name,

JosseVanDelm: Yeah that's a bit silly, I propose we make the barrier enable (if required for later editions, an accelerator parameter instead, then it will also be set just once)

compiler/transforms/convert_acc_to_csr.py
JosseVanDelm: No we should totally get rid of this, the NOPs in this routine are similarly weird and stupid, but it's a dontfix typ of situation for the HWPE accelerator.
If it makes sense we can ask the hardware people to fix it in SNAX

JosseVanDelm:
@AntonLydike I've looked over all HWPE addresses and have adapted them accordingly.
I'm also emitting the poll-style barrier as an SCF while directly now instead of trying to inline multiple lines of assembly, it looks like MLIR doesn't like that.

I propose to switch to a single barrier (and to not have a barrier_enable property right now in the acceleratorOp)
I think it makes more sense to add the barrier_enable to the actual setup, since it also has to be setup just once.
I can create an "interface"/"abc" for properly emitting barriers in a later step when we put all of this stuff for accelerators in together in a big refactoring. wdyt?

JosseVanDelm:
Okay so I found a bunch of errors I introduced :)

AntonLydike:
Hmm, I thought a bit about ideas on how to handle barriers and stuff. Maybe it makes sense to have a "how to barrier" region on the accelerator op, that basically constains arbitrary operations that express how barriers are used, and that can be inlined in the lowering. This feels a bit like IR abuse, but I'd be up for trying it.

Essentially this would look like this:

```mlir
acc2.accelerator {...} await {
  scf.while {
    %cond = llvm.inline_asm "csrr ..." : () -> i1
    scf.yield %cond
  } do {}
}

// ...

acc2.await(%token) ...
```

becomes after lowering:

```mlir
// acc2.accelerator op deleted

// ...

// the await part is just inlined to replace the await op:
  scf.while {
    %cond = llvm.inline_asm "csrr ..." : () -> i1
    scf.yield %cond
  } do {}
```

JosseVanDelm:
@AntonLydike what would be the benefit of declaring the barriers inside the IR?
Because right now there are two types of barriers we support, which in my opinion is already one too many :)
I'd like to keep stuff as easy/dumb as possible.

Do you think it is essential that we add support for this in this PR?

JosseVanDelm:
> Okay so I found a bunch of errors I introduced :)

I have unmarked the previously introduced test from XFAIL to address the previous issue :)

@jorendumoulin ready for your review!

compiler/dialects/acc.py
```diff
@@ -189,12 +212,84 @@ def verify_(self) -> None:
...
+class AcceleratorSymbolOpTrait(SymbolOpInterface):
+    def get_sym_attr_name(self, op: Operation) -> StringAttr | None:
+        assert isinstance(op, AcceleratorOp)
+        return StringAttr(op.name_prop.string_value())
+
+
+@irdl_op_definition
+class AcceleratorOp(IRDLOperation):
+    """
+    Declares an accelerator that can be configures, launched, etc.
```
jorendumoulin, line 224:
```suggestion
    Declares an accelerator that can be configured, launched, etc.
```

compiler/dialects/acc.py
```diff
@@ -189,12 +212,84 @@ def verify_(self) -> None:
...
+            fields = DictionaryAttr(
+                {name: IntegerAttr(val, i32) for name, val in fields.items()}
+            )
+
+        super().__init__(
+            properties={
+                "name": (
+                    SymbolRefAttr(name) if not isinstance(name, SymbolRefAttr) else name
+                ),
+                "fields": fields,
```
jorendumoulin, line 258: fields is accepted as both dict and `DictionaryAttr`, does this also need to be checked here?

compiler/transforms/convert_acc_to_csr.py
```diff
@@ -0,0 +1,273 @@
...
+                llvm.InlineAsmOp(
+                    "csrw $0, $1",
+                    "I, r",
+                    [addr_val, zero],
+                    has_side_effects=True,
+                ),
+                # Three nops for random but important reasons
+                llvm.InlineAsmOp("nop", "", [], [], has_side_effects=True),
+                llvm.InlineAsmOp("nop", "", [], [], has_side_effects=True),
+                llvm.InlineAsmOp("nop", "", [], [], has_side_effects=True),
```
jorendumoulin, line 164: what is the reason here?

compiler/transforms/convert_acc_to_csr.py
```diff
@@ -0,0 +1,273 @@
...
+        for field, val in op.iter_params():
+            addr = field_to_csr[field]
+            rewriter.insert_op_before_matched_op(
+                [
+                    addr_val := arith.Constant(addr),
+                    llvm.InlineAsmOp(
+                        "csrw $0, $1",
+                        "I, r",
+                        [addr_val, val],
+                        has_side_effects=True,
```
jorendumoulin, line 73: Can we do the "I, K" constraint here too? Or is llvm not smart enough to decide which is best?

jorendumoulin (approved):
This looks good!

compiler/transforms/convert_acc_to_csr.py
JosseVanDelm: I don't know hahaha @rgantonio ?
Code in question is here: https://github.com/KULeuven-MICAS/snitch_cluster/blob/8d7dff1f33fd9705bfcf1bbcebba246a5ccac25d/target/snitch_cluster/sw/snax/mac/src/mac.c#L16

compiler/transforms/convert_acc_to_csr.py
JosseVanDelm: LLVM is smart!

PR #111, opened 2024-03-23
use offset in memref-to-snax

jorendumoulin:
The offset was not considered when calculating the size to allocate for TSL layouts. This PR fixes that


JosseVanDelm (approved):
Yes

PR #115, opened 2024-04-16
acc2: Add an accelerator registry for lowering transforms

JosseVanDelm:
This PR solves the problem that if there are multiple accelerators in the IR they should be lowered in a different way.
It adds a "registry" = a dict that maps symbol names to the Accelerator interface introduced in #114 .
That way we can query the right lowering through the `accelerator.op`'s symbol at runtime.

Right now it's very simple, but this can be made more complex in the future if necessary.

Note that this requires the IR to be annotated with the acc.AcceleratorOp, otherwise the right passes can not be queried. Before they were hardcoded :) so it wasn't necessary to put them there.
Note that the lowerings from linalg are still hardcoded. This is work for the next PR.

### DIFF
--- a/compiler/transforms/convert_acc_to_csr.py
+++ b/compiler/transforms/convert_acc_to_csr.py
@@ -17,7 +17,6 @@ from xdsl.traits import SymbolTable
 
 from compiler.accelerators.accelerator import Accelerator
 from compiler.accelerators.registry import get_registered_accelerators
-from compiler.accelerators.snax_hwpe_mult import SNAXHWPEMultAccelerator
 from compiler.dialects import acc
 
 
@@ -191,15 +190,12 @@ class RemoveAcceleratorOps(RewritePattern):
         rewriter.erase_matched_op()
 
 
-@dataclass(frozen=True)
 class ConvertAccToCsrPass(ModulePass):
     """
     Converts acc2 dialect ops to series of SNAX-like csr sets.
     """
 
     name = "convert-acc-to-csr"
-
-    accelerator: type[Accelerator] = SNAXHWPEMultAccelerator
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
         # first lower all acc2 ops and erase old SSA values
--- a/compiler/transforms/convert_acc_to_csr.py
+++ b/compiler/transforms/convert_acc_to_csr.py
@@ -43,10 +43,13 @@ class LowerAccBasePattern(RewritePattern, ABC):
         acc_op = trait.lookup_symbol(self.module, accelerator)
         if not isinstance(acc_op, acc.AcceleratorOp):
             raise RuntimeError(
-                f"Invalid IR: no accelerator op for @{accelerator.data} found in module"
+                f"Invalid IR: Lowering acc2 for accelerator '{accelerator.data}'"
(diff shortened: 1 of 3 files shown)


JosseVanDelm:
Okay something's off, I still have to adapt the linalg-to-acc pass, but it should still work?

compiler/transforms/convert_acc_to_csr.py
```diff
@@ -41,7 +45,9 @@ def get_acc(self, accelerator: StringAttr) -> acc.AcceleratorOp:
             raise RuntimeError(
                 f"Invalid IR: no accelerator op for @{accelerator.data} found in module"
             )
-        return acc_op
+        # Get accelerator interface with symbolref without @
+        acc_info = get_registered_accelerators()[str(acc_op.name_prop)[1:]]
```
jorendumoulin, line 49: can we emit a more informative error if the accelerator in the module is not registered?

compiler/transforms/convert_acc_to_csr.py
```diff
@@ -57,12 +63,10 @@ class LowerAccSetupToCsr(LowerAccBasePattern):
...
+        acc_op, acc_info = self.get_acc(op.accelerator)
         # grab a dict that translates field names to CSR addresses:
-        acc_info = SNAXHWPEMultAccelerator()
         # emit the llvm assembly code to set csr values:
-        rewriter.insert_op_before_matched_op(
-            acc_info.lower_acc_setup(op, self.get_acc(op.accelerator))
-        )
+        rewriter.insert_op_before_matched_op(acc_info.lower_acc_setup(op, acc_op))
         # delete the old setup op
         rewriter.erase_matched_op(safe_erase=False)
```
jorendumoulin, line 71: why not `rewriter.replace_matched_op()` ?

jorendumoulin (commented):
I love where this is going, this will be super useful!

compiler/transforms/convert_acc_to_csr.py
JosseVanDelm: Adapted!

compiler/transforms/convert_acc_to_csr.py
```diff
@@ -39,9 +42,14 @@ def get_acc(self, accelerator: StringAttr) -> acc.AcceleratorOp:
...
             raise RuntimeError(
-                f"Invalid IR: no accelerator op for @{accelerator.data} found in module"
+                f"Invalid IR: Lowering acc2 for accelerator '{accelerator.data}'"
+                " requires an acc2.accelerator op to declare a symbol for "
+                f"@{accelerator.data}, but no such symbol was found"
+                " in the current module."
             )
-        return acc_op
+        # Use the retrieved acc_op to retrieve information from the registry
+        acc_info = get_registered_accelerators()[acc_op.name_prop.string_value()]
```
jorendumoulin, line 51: Actually this is what i meant: otherwise it will be an ugly keyerror or something
```suggestion
        if acc_op.name_prop.string_value() not in get_registered_accelerators():
            raise RunTimeError(f"Unknown accelerator symbol {acc_op.name_prop.string_value()}")
        acc_info = get_registered_accelerators()[acc_op.name_prop.string_value()]
```

jorendumoulin (approved):
Great!

PR #117, opened 2024-04-17
acc2: Add registry for converting to acc2 dialect

JosseVanDelm:
This adds the remaining part that needs to be queried from the AcceleratorRegistry to go to the acc2 dialect.

### DIFF
--- a/compiler/accelerators/accelerator.py
+++ b/compiler/accelerators/accelerator.py
@@ -11,6 +11,9 @@ class Accelerator(ABC):
     Interface to lower to and from acc2 dialect.
     """
 
+    name: str
+    fields: tuple[str]
+
     @abstractmethod
     def generate_setup_vals(
         self, op: Operation
--- a/compiler/accelerators/registry.py
+++ b/compiler/accelerators/registry.py
@@ -1,11 +1,26 @@
+from collections.abc import Iterable, Mapping
+
 from compiler.accelerators.accelerator import Accelerator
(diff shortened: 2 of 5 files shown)


JosseVanDelm:
I need to add a pass that inserts the missing accelerator info for the snax_hwpe_mult still. Hence the failing tests

compiler/transforms/insert_acc_op.py
```diff
@@ -0,0 +1,63 @@
...
+    def match_and_rewrite(self, op: builtin.ModuleOp, rewriter: PatternRewriter):
+        # Get module symbol table and attach accelerator op to it
+        t = op.get_trait(SymbolTable)
+        assert t is not None
+        acc_string = self.acc_op.name_prop.string_value()
+        if t.lookup_symbol(op, acc_string) is not None:
+            raise RuntimeError(
+                f"Cannot insert op: @{acc_string}, already present in symbol table."
+            )
+        t.insert_or_update(op, self.acc_op)
```
jorendumoulin, line 37: does adding something to the symbol table also put it in the IR?

jorendumoulin: aha, now i see the test. nice

jorendumoulin (approved):
lovely

compiler/transforms/insert_acc_op.py
JosseVanDelm: Apparently it does hahaha

PR #118, opened 2024-04-19
Add support for Gemmini

JosseVanDelm:
Adds a gemmini accelerator that implements a RoCC interface.

RoCC in Gemmini is a bit special, it uses 2 source registers (and hence consumes 2 SSAValues).
This is not really supported in the current flow (Dedup will remove them not taking into account if they need to be set together or not). 

Currently the flow assumes that all fields are set with an rs2 and rs1 field, otherwise it will fail (see attached XFAIL test).
I'm going to add proper conversions for deduplicated code later.

There are some other issues to be solved later:
* ~~What to do with launch-semantic configuration ops? These are not really launch ops, neither are they setup ops.~~
This is now implemented as per #119 .
* Actually run this thing in spike
* ...?

PR #119, opened 2024-04-23
acc2: Add values to launch field

JosseVanDelm:
This PR adds the possibility of adding SSAValues and fields to the acc2.LaunchOp.
This is to fix the case where some accelerators also perform some setup implicit on each launch.
The nice thing is that encoding it like this immediatly makes it impossible to deduplicate those calls (that never have to be deduplicated in the first place) but still allows us to perform some configuration inside the launching of accelerators.
This is necessary to be pushed first to allow for #118 to be added

This also replaces the generate_setup_vals with an even more opaque mechanism in the interface to generate the entire sequence of setup, launch and await ops instead, which is required in this case, since some setup operations are not tied to just the setup op anymore

### DIFF
--- a/compiler/accelerators/accelerator.py
+++ b/compiler/accelerators/accelerator.py
@@ -35,7 +35,7 @@ class Accelerator(ABC):
         "acc2.accelerator"() <{
             name            = @name_of_the_accelerator,
             fields          = {field_1=address_1, field_2=address2},
-            launch_addr     = launch_address,
+            launch_fields   = {launch_field_1=address_1,
             barrier         = barrier_address,
         }> : () -> ()
         """
--- a/compiler/accelerators/accelerator.py
+++ b/compiler/accelerators/accelerator.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 
-from xdsl.ir import Operation, SSAValue
+from xdsl.ir import Operation
 
 from compiler.dialects import acc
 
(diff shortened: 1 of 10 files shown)


jorendumoulin (commented):
Looks good, but I have one important question: right now it seems we allow for multiple fields for the launch op. Do we want this? I imagined there would still be one launch op (with no need to specify the "launch"key. Is this because we would maybe need to setup multiple values in one launch op? (Gemmini style)

JosseVanDelm:
Hi @jorendumoulin,

It is indeed written with the purpose of attaching multiple values to the launch op.
With the current implementation of SNAX it indeed looks a bit like overkill, since there's just one value (technically it doesn't even need to be passed to the launch op, since it is always the same value for SNAX (zero).
So it can just be inserted by the conversion pass that lowers the launch ops.
In this PR I moved the place where that zero is inserted to the lowering of the linalg op, since that would be required if we lower something like gemmini, and testing whether it works here is actually simpler than with gemmini.

I can imagine there can be multiple things that need to be setup (it is the case in gemmini) and where those values are not always the same (also the case for gemmini).

jorendumoulin:
Alright, now it just looks weird as in this case it is possible to launch the accelerator with 2 different csr addresses, which I don't think is the case. I guess this awaits the implementation of the system for multiple values per CSR. 

In that case, I don't think it should be possible to specify multiple csr addresses for the launch (should it?). For it's fine to solve this issue later when it is clear how multiple values per CSR are handled.

jorendumoulin (approved):

PR #120, opened 2024-05-02
acc2: Add deduplication recovery support for gemmini 

JosseVanDelm:
Note: stacked PR

This adds support to emit RoCC instructions for setup values that were previously deduplicated.
In certain cases this means that you need to trace back what the original rs1 was associated to an rs2 or vice versa.

Note that this changes a previously implemented test to not be failing anymore.

### DIFF
--- a/compiler/accelerators/rocc.py
+++ b/compiler/accelerators/rocc.py
@@ -1,11 +1,13 @@
 from abc import ABC
-from collections.abc import Sequence
+from collections.abc import Iterable, Sequence
 
 from xdsl.dialects import llvm
+from xdsl.dialects.builtin import IntegerAttr
 from xdsl.ir import Operation, SSAValue
 
 from compiler.accelerators.accelerator import Accelerator
 from compiler.dialects import acc
+from compiler.inference.trace_acc_state import infer_state_of
 
 
 class RoCCAccelerator(Accelerator, ABC):
@@ -26,7 +28,7 @@ class RoCCAccelerator(Accelerator, ABC):
         launch_op: acc.LaunchOp, acc_op: acc.AcceleratorOp
     ) -> Sequence[Operation]:
         xcustom_acc = 3  # hardcoded to 3 for now
-        vals = create_pairs(launch_op, acc_op.launch_field_names())
+        vals = create_pairs(launch_op)
         # Create the sequence of all operations that need to be emitted
         return combine_pairs_to_ops(acc_op.launch_field_items(), vals, xcustom_acc)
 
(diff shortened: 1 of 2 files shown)


compiler/accelerators/rocc.py
```diff
@@ -35,52 +37,84 @@ def lower_acc_setup(
...
+    # For launch_ops, no tracing back is necessary, since dedup doesn't happen
+    elif isinstance(fields_op, acc.LaunchOp):
+        pass
+    # Assert that pairs of rs1 and rs2 exist for each item in the fields
+    for instruction in instructions:
+        assert instruction + ".rs1" in field_dict, f"No rs1 found for {instruction}"
+        assert instruction + ".rs2" in field_dict, f"No rs2 found for {instruction}"
+    # Create a dictionary that contains the two vals associated
+    # to each single RoCC instruction
+    map: dict[str, tuple[SSAValue, SSAValue]] = {}
```
jorendumoulin, line 81: can we change the variable name as to not confuse it with https://www.geeksforgeeks.org/python-map-function/

compiler/accelerators/rocc.py
```diff
@@ -35,52 +37,84 @@ def lower_acc_setup(
...
+) -> dict[str, tuple[SSAValue, SSAValue]]:
+    """
+    For a given RoCC launch or setup op, return a map that maps a single RoCC
+    instruction to pairs of two SSAValues
+
+    For setup ops, this can retrace back previous setup state if necessary
+    (i.e. if one of the two operands of an instruction gets dedupped)
+    """
+    # Make a set of all the unique instruction names in the current operation
+    instructions = set([name[:-4:] for name, _ in fields_op.iter_params()])
```
jorendumoulin, line 61: what does this `name[:-4:]` wizardry do?

jorendumoulin: wait, is this just the same as `name[:-4]`? so just removing the last part (.rsx)? this would make sense, but the second colon confused me 🙃

jorendumoulin (approved):
Looks good! But my general feeling is that the system for one instruction for setting up multiple values is quite patched on here.

A more structural approach seems quite a challenge to design, but i think it would lead to simpler code + less duplication of logic if the same system would occur someplace else than gemmini. 
Just something to keep in mind.

compiler/accelerators/rocc.py
jorendumoulin: I'm really not a fan of string parsing for logic like this, although I don't see a better solution for this case right now

JosseVanDelm: @jorendumoulin the current system is quite brittle indeed :(

> although I don't see a better solution for this case right now
Same!

I've been wondering how I can add a verifier inside the interface to check for some things: i.e.
Making sure there is a rs1 and rs2 behind everything, making sure all of those pairs are pointing to the same address, but I'm not sure where to put those (making the acceleratorop verifier call into the acceleratorinterface seems like a no-no?)

JosseVanDelm:
@jorendumoulin yes I completely agree.
I've been thinking of ways to support it better, but until we get there, this should be an okay example to have in there :smile:

PR #122, opened 2024-05-06
make: Fix clean for mlperf tiny

jorendumoulin:
this pr sets up a proper clean statement in the makefiles for the mlperf tiny benchmarks


JosseVanDelm (approved):
I Like it better with this commit message but otherwise fine!

PR #124, opened 2024-05-07
acc2: Add noimplemented to ABC

JosseVanDelm:
Seems like the proper way to define abstract base classes is to throw errors if they are ever used instead of just passing.


jorendumoulin (approved):
huh

PR #126, opened 2024-05-15
Add stuff to gitignore

AntonLydike:
I'm just copying some stuff from xDSL so that I have less pain with git :smile:


JosseVanDelm (approved):
Sure!

PR #128, opened 2024-05-16
ruff: increase line length

jorendumoulin:
because anton is a modern man with a modern terminal

is 120 enough? 300 like you have in xDSL seems outrageous


AntonLydike (approved):
*Back to the Future noises*

:heart:

PR #130, opened 2024-05-28
docker: Use monorepo docker from SNAX repository instead

JosseVanDelm:
Using https://github.com/KULeuven-MICAS/snax_cluster/pull/138/ to test if this prebuilt docker works.


JosseVanDelm:
It appears https://github.com/KULeuven-MICAS/snax_cluster/pull/138/commits/34aa29c750830f83264be28d9d57bae26c51b87e finally did the trick

jorendumoulin:
This is now running successfully on the docker image of kuleuven-micas/snax_cluster#146

I removed the custom toolchain again because we really don't need it :) just remove the includes from the makefile. Those linker includes are only necessary when compiling with non-generic llvm toolchain, which we haven't done for quite a while now

JosseVanDelm:
@jorendumoulin lol

JosseVanDelm:
@jorendumoulin feel free to approve this.
@xiaoling-yi  @rgantonio could you please let us know if you have understood the consequences of us moving to a monolithic container for the compiler tests as well? It means all of our work inside docker is directly affected by changes in the snax_cluster repository (w.r.t. the dockerfile), but it also means everyone is using the same versions of everything, which should make integrating things easier.

jorendumoulin (approved):
Let's go! We just have to set the docket tags back to main

rgantonio:
@JosseVanDelm I mean yes but then that means we also need to keep track if we break stuff in the `snax-mlir`?

So for example, say it passes in `snax_cluster` but then fails on `snax-mlir`? Though ideally if it passes in `snax_cluster` it should pass on your side right?

JosseVanDelm:
@jorendumoulin this is a good comment from Ryan, maybe we should still put the prebuilts in our specifc version of the container? We should be able to pin a version of the prebuilts, which now we can not do :(

jorendumoulin:
As a quickfix, we can use the docker hash.

Maybe a better solution is to start versioning the snax cluster (v0.1.0), and use those as references? What do you think about this @rgantonio @JosseVanDelm

rgantonio:
The versioning is a good idea! So that you're not too reliant on every change we make.

Because right now it's like there are several changes that @xiaoling-yi wants to do with the streamers. Those that need to get updated according to what Marian sees too.

JosseVanDelm:
Quickfix = ok
Real versioning = better

PR #129, opened 2024-05-27
accfg: Loop hoisting for scf.for loops

AntonLydike:
- [x] Initial implementation
- [x] Clean up code, add comments
- [x] Add test for nested loops
- [x] Add test for multi-accelerator loops
- [x] Add end-to-end test 
- [x] Add test for multi-setup-ops loop

PR #131, opened 2024-05-28
Add custom syntax for acc.setup

AntonLydike:
Broke this out of #129 because it's already a chonker

This makes it easier to understand setup ops (e.g. the input state is no longer mixed with the values, and the "key = value" pairs are closer together.


JosseVanDelm (approved):
I love this. I think we can iterate on the actual format a bit more though!

PR #134, opened 2024-05-29
Add DMA barrier hotfix

JosseVanDelm:
There was a missing barrier in one of the runtime routines, resulting in fun-to-debug erronous behaviour.


AntonLydike (approved):
Most curious this one

PR #139, opened 2024-06-06
Add `accfg-config-overlap` pass

AntonLydike:
Very simple pass for now, just moving inside blocks. Maybe next up something for for loops, idk.


flake.nix
```diff
@@ -0,0 +1,28 @@
...
+              LD_LIBRARY_PATH = "${stdenv.cc.cc.lib}/lib";
+              buildInputs = [
+                nodejs_22 ruff black isort llvmPackages_17.mlir
+              ];
+	      name = "snax-nix";
+              JAVA_TOOL_OPTIONS = "-Didea.config.path=${conf_basebath}/conf -Didea.system.path=${conf_basebath}/sys -Didea.plugins.path=${conf_basebath}/plugins -Didea.log.path=${conf_basebath}/sys/logs";
+            };
+          }
+    );
+}
```
AntonLydike, line 28: This just makes my life easier with NixOS. Basically installs nodejs, mlir-opt-17 and some other things, and tells jetbrains to start an isolated instance so that I get the correct environment variables inherited.

compiler/transforms/accfg_config_overlap.py
```diff
@@ -0,0 +1,74 @@
...
+    Inspects a single block for the following structure:
+
+    ```
+     %token = accfg.launch %state ...
+     // ...
+     %next_state = accfg.setup ... from %state ...
+    ```
+
+    And moves the setup op right behind the launch, iff:
+    - The these two ops are the only two uses of the `%state` SSA value
```
JosseVanDelm, line 25:
```suggestion
    - These two ops are the only two uses of the `%state` SSA value
```

compiler/transforms/accfg_config_overlap.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: accfg.SetupOp, rewriter: PatternRewriter, /):
+        # ignore setup without an input state
+        if not op.in_state:
+            return
+        # only apply if there are exactly two uses of the state
```
JosseVanDelm, line 33: This is to not fiddle with the order of the a launch op that has been setup by a fully deduplicated setup?
Or why is this check here?

compiler/transforms/accfg_config_overlap.py
```diff
@@ -0,0 +1,74 @@
...
+        else:
+            return
+
+        # check that the launch and the setup are on the same block level
+        if launch.parent_block() != op.parent_block():
+            return
+
+        # abort if the launch op has already been moved (prevent forever loops)
+        if launch.next_op is op:
+            return
```
JosseVanDelm, line 54: Maybe it makes more sense to put this on the top as the first check?

JosseVanDelm (changes requested):
This pass looks good.

I'm not a big fan of the nix lock file though.
I understand this makes your life easier, but I don't understand why we need this inside this repository? It also seems quite specific to your specific setup with your IDE?
Regardless, it should not be part of this PR IMHO.

compiler/transforms/accfg_config_overlap.py
AntonLydike: We can only do this after we have the reference to the `launch`, so can only move it up one check, but I don't think that matters much.

compiler/transforms/accfg_config_overlap.py
AntonLydike: (basically, if there is more than one use, I don't see a point on figuring out what the "last" move is, and it makes all other checks more complicated)

compiler/transforms/accfg_config_overlap.py
JosseVanDelm: Right! Ok!

JosseVanDelm (approved):
Cool! Thanks @AntonLydike !

PR #140, opened 2024-06-06
[WIP] Add support for `scf.for` in accfg-config-overlap

AntonLydike:
Still WIP.


AntonLydike:
I will split out all the refactoring and the fact that I re-did the block level moves!

compiler/transforms/accfg_config_overlap.py
```diff
@@ -0,0 +1,214 @@
...
+
+    ```
+     %token = accfg.launch %state ...
+     // ...
+     %next_state = accfg.setup ... from %state ...
+    ```
+
+    And moves the setup op right behind the launch, iff:
+    - These two ops are the only two uses of the `%state` SSA value
+    """
```
AntonLydike, line 31: This needs clarification from me, it's no longer accurate!

AntonLydike:
No longer WIP!

jorendumoulin (approved):
This is very nice and super clean!

I think I understand most of it 😄 , but I don't have any significant comments.
Does this handle the following case where you use a value computed using the loop index `i`? 
I think this is handled by your helper functions but I wanna make sure
```
for i = 0 .. N
    setup (2*i, 4*i)
    await
```

AntonLydike:
> This is very nice and super clean!
> 
> I think I understand most of it 😄 , but I don't have any significant comments. Does this handle the following case where you use a value computed using the loop index `i`? I think this is handled by your helper functions but I wanna make sure
> 
> ```
> for i = 0 .. N
>     setup (2*i, 4*i)
>     await
> ```

Good catch! I shall add filechecks for this so that we can actually show that it works :smile:

jorendumoulin:
lovely!

compiler/tools/snax_opt_main.py
```diff
@@ -46,7 +46,7 @@ def __init__(
 
         ## Add custom dialects & passes
         # FIXME: override upstream accfg dialect. Remove this after upstreaming full downstream accfg dialect.
-        del self.ctx._registered_dialects["accfg"]  # pyright: ignore
+        self.ctx._registered_dialects.pop("accfg", None)
```
AntonLydike, line 49: This change is so that if an xdsl is used that does *not* have an `accfg` dialect, we don't get an error saying "hurr durr no "accfg" in `_registered_dialects`....

compiler/transforms/accfg_config_overlap.py
```diff
@@ -84,6 +85,130 @@ def match_and_rewrite(self, op: accfg.SetupOp, rewriter: PatternRewriter, /):
...
+    3. We insert a copy of the setup af the end of the loop, replacing dependencies with the next iterations variables
+       (which we get from the yield / by adding %step to %i)
+    4. We erase the original setup op.
+
+
+    This results in the following IR post-optimization:
+
+    ```
+     %s0 = setup ... // loop-initial state
+     %l1 = setup from %l0 to ("i" = %lb)
```
JosseVanDelm, line 114: This seems incorrect
```suggestion
     %l1 = setup from %s0 to ("i" = %lb)
```

compiler/transforms/accfg_config_overlap.py
```diff
@@ -84,6 +85,130 @@ def match_and_rewrite(self, op: accfg.SetupOp, rewriter: PatternRewriter, /):
...
+     scf.for (%i = %lb to %ub step %step) iter_args(%l0 = %s0) ... {
+       %t = launch(%l0)
+       await(%t)
+       %i_next = arith.addi %i, %step
+       %l1 = setup from %l0 to ("i" = %i_next)
+       yield %l1
+     }
+    ```
+    The setup will then be lifted before the await by teh BlockLevelSetupAwaitOverlapPattern.
```
JosseVanDelm, line 123:
```suggestion
    The setup will then be lifted before the await by the BlockLevelSetupAwaitOverlapPattern.
```

tests/filecheck/transforms/accfg-config-overlap.mlir
JosseVanDelm, line 1: what happens if there are multiple setup ops in the loop body?
Can you add a test for that case?
```
for
   for
        setup a= 2 c=3
        launch
        await
        setup b= 3 c=3
        launch
        await
```
will it do:
```
setup c=3 a=2
for
   for
        launch
        setup b= 3
        await
        launch
        setup a = 2
        await
```

JosseVanDelm (approved):
This is wild

tests/filecheck/transforms/accfg-config-overlap.mlir
AntonLydike: it should^^ let's see! I haven't thought about that case hard enough yet!

AntonLydike: Tests added!

tests/filecheck/transforms/accfg-config-overlap.mlir
```diff
@@ -82,3 +82,177 @@ func.func @simple_negative(%A: i32, %B: i32, %i1: i1) {
...
+
+  %1 = scf.for %i = %lb to %ub step %step iter_args(%l0 = %0) -> (!accfg.state<"simple">) : i32 {
+
+    // first setup, sets up %A
+    %i_plus_2 = arith.addi %i, %c2 : i32
+    %l1 = accfg.setup "simple" from %l0 to ("A" = %A : i32, "B" = %A : i32, "i" = %i_plus_2 : i32) : !accfg.state<"simple">
+    %t1 = "accfg.launch"(%l1) <{param_names = [], accelerator = "simple"}> : (!accfg.state<"simple">) -> !accfg.token<"simple">
+    "accfg.await"(%t1) : (!accfg.token<"simple">) -> ()
+
+    // second setup, sets ub %B
```
JosseVanDelm, line 216:
```suggestion
    // second setup, sets up %B
```

JosseVanDelm (approved):
Awesome work @AntonLydike !

PR #144, opened 2024-06-11
Add ScopedSetupWithInputs (stacked 2/3)

AntonLydike:
This PR adds ScopedSetupWithInputs and moves a bunch of helpers out of the rewrites and into the `inference` module.

ScopedSetupWithInputs represents a setup op together with operations that calculate it's inputs. These operations must be pure, and be contained within a block (usually the block at which the operation is located).

This helper class is needed for stacked PR 3/3, and will also be helpful to simplify some other rewrites I'm sure,

### DIFF
--- a/compiler/inference/helpers.py
+++ b/compiler/inference/helpers.py
@@ -1,0 +1,141 @@
+from collections.abc import Generator
+
+from xdsl.dialects import scf
+from xdsl.ir import Block, BlockArgument, Operation, OpResult, Region, SSAValue
+
+from compiler.dialects import accfg
+
+
+def get_initial_value_for_scf_for_lcv(loop: scf.For, var: SSAValue) -> SSAValue:
+    """
+    Given a loop-carried variable inside an scf for loop as the block argument,
+    return the SSA value that is passed as the initial value to it.
+    """
+    if var not in loop.body.block.args:
+        raise ValueError(
+            f"Given value {var} not a block argument of the for loop {loop}!"
+        )
+    idx = loop.body.block.args.index(var) - 1
+    return loop.iter_args[idx]
+
+
+def val_is_defined_in_block(val: SSAValue, block: Block) -> bool:
+    """
+    Check if val is defined in block or in blocks nested in block.
+    """
+    if isinstance(val, BlockArgument):
+        block_ptr: Block | None = val.owner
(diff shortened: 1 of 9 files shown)


compiler/inference/helpers.py
AntonLydike, line 1: Nothin in this file is new, it's just stuff moved out of the specific rewrites in an attempt to clean up.

compiler/inference/helpers.py
```diff
@@ -0,0 +1,141 @@
...
+    if end_op is not None:
+        assert (
+            start_op.parent_block() is end_op.parent_block()
+        ), "Cannot iterate between operations not within the same block!"
+
+    yield start_op
+    while start_op.next_op is not end_op:
+        assert start_op.next_op is not None  # just for pyright
+        yield start_op.next_op
+        start_op = start_op.next_op
```
AntonLydike, line 141: This is a neat helper that I might upstream to xDSL at some point. It's new in this PR

compiler/transforms/accfg_config_overlap.py
```diff
@@ -53,10 +57,30 @@ def match_and_rewrite(self, op: accfg.SetupOp, rewriter: PatternRewriter, /):
...
+        if all(
+            between_op in inputs.inputs for between_op in iter_ops_range(launch, op)
+        ):
+            return
+
+        inputs.lazy_move_up(  # and move the setup and inputs to be right behind the launch
+            parent_block,
+            InsertPoint.after(launch),
+            rewriter,
+        )
```
JosseVanDelm, line 83: Maybe we should make clear that this is a heuristic? We can move these ops anywhere in the block iiuc, and not all possible move strategies might make as much sense depending on the scenario?

JosseVanDelm (approved):
I've added a few comments:
It seems like this is implemented for scf.for ops (in the comments), but the code seems to cover more stuff? Can you clarify/address those comments?
Apart from that this looks very neat!

Can you have a look at my heuristic comment and tell me what you think about it?

AntonLydike:
Thanks for the feedback!

compiler/transforms/accfg_config_overlap.py
AntonLydike: Not sure what you mean exactly by this...

PR #146, opened 2024-06-12
support strided copies

jorendumoulin:
This PR implements support for copying strided memrefs, including those with transformations. This works by creating a TSL representation of the stridedlayoutattr and then applying the existing algorithm on the tsl representations.


JosseVanDelm (approved):
Maybe at some point we should show the explicit conversions to the layout attributes in the IR, and split this pass into two different passes? Now this conversion is a bit hidden, would be better to see this upfront

PR #147, opened 2024-06-17
fix strided transform copy bug

jorendumoulin:
Apply element strides for strided attribute when forming tsl from `StridedLayoutAttr`, as this accounts for element width, while tsl does not


compiler/transforms/snax_copy_to_dma.py
```diff
@@ -109,7 +109,10 @@ def extract_strides(memreftype: MemRefType):
     """
     if isinstance(memreftype.layout, StridedLayoutAttr):
         strides = memreftype.layout.strides.data
-        strides = [x.data if isinstance(x, IntAttr) else None for x in strides]
+        element_stride = memreftype.element_type.width.data // 8
```
JosseVanDelm, line 112:
```suggestion
        element_stride = memreftype.element_type.width.data // 8 # bits to bytes
```

JosseVanDelm (approved):
Good for me, modulo failing tests

PR #150, opened 2024-06-18
bump snax version

jorendumoulin:


jorendumoulin:
I had to fix the numpy version as its first major release since 2006 came out a couple of days ago, which breaks some of the passes

container/Dockerfile
JosseVanDelm, line 1: This one is not used, is it?

JosseVanDelm (approved):

container/Dockerfile
jorendumoulin: I created #152

PR #155, opened 2024-06-18
Where we are going we don't need safety

AntonLydike:
Notices a the new rewriting behaviour to fail when erasing state variables in loops, so now I made our rewrite tell the rewriter that we are in *expert mode*.


JosseVanDelm:
Where are you noticing this behaviour? I don't understand what the safe thing is doing now

JosseVanDelm (approved):
I see. I guess the old tests are still running which means this is okay to merge?

PR #157, opened 2024-06-19
bump snax to v0.1.3

jorendumoulin:
to get the hardware fix of kuleuven-micas/snax_cluster#163


JosseVanDelm (approved):
Can we also get the newer alu pls

PR #160, opened 2024-06-24
Adapt sizes to new 64 bit pointers in streamer alu

JosseVanDelm:
I've corrected the pointers to not include the sign extension anymore and to also not use magic pointer arithmetic (to avoid future confusion)

### DIFF
--- a/kernels/streamer_alu/main.c
+++ b/kernels/streamer_alu/main.c
@@ -3,27 +3,27 @@
 #include "snax_rt.h"
 #include "snrt.h"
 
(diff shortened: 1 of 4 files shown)


kernels/streamer_alu/main.c
```diff
@@ -65,7 +63,5 @@ int main() {
 
   printf("Accelerator Done! \n");
   printf("Accelerator Cycles: %d \n", perf_count);
-  printf("Number of errors: %d \n", err);
-
```
JosseVanDelm, line 69: i've also removed this since it is printed in the simulator itself

jorendumoulin (approved):
Thanks! This is probably better than expert mode haha

PR #164, opened 2024-06-26
Add snax.mcycle op

JosseVanDelm:
This PR adds an operation that adds a utility mcycle operation to the snax dialect.
This is useful to split up traces in multiple sections

### DIFF
--- a/compiler/dialects/snax.py
+++ b/compiler/dialects/snax.py
@@ -24,7 +24,6 @@ from xdsl.irdl import (
     result_def,
     var_operand_def,
 )
-from xdsl.traits import Pure
 from xdsl.utils.exceptions import VerifyException
 
 from compiler.util.memref_descriptor import LLVMMemrefDescriptor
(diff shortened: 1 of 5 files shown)


AntonLydike:
I wonder what it would take to re-use the lowerings already in xdsl for this?

AntonLydike:
After talking to Sasha, it does not. They implement it in their C wrapper code.

AntonLydike (approved):
This feels a bit too hacky even for my taste, but I guess the show must go on...

JosseVanDelm:
> I wonder what it would take to re-use the lowerings already in xdsl for this?

Oops I did not consider this.

> This feels a bit too hacky even for my taste, but I guess the show must go on...

Not sure what you mean by this. Calling into C code for this purpose seems a lot hackier in my opinion?
I guess right now it is a bit basic, but it should serve it's purpose just fine I think?

PR #166, opened 2024-06-27
dialects: (accfg) Add accfg.effects attribute

AntonLydike:
The idea is that:

- function calls (`func.call` and `llvm.call`) are pessimistically assumed to effect the accelerator
- any other op does not effect the accelerator if ops in their body don't 
- Both of these assumptions can be overwritten by setting the `accfg_effects` attribute to `#accfg.effects<none>` or `#accfg.effects<full>`.

### DIFF
--- a/compiler/dialects/accfg.py
+++ b/compiler/dialects/accfg.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Iterable
+from enum import Enum
 
 from xdsl.dialects.builtin import (
     ArrayAttr,
@@ -12,6 +13,7 @@ from xdsl.dialects.builtin import (
 )
 from xdsl.ir import (
     Attribute,
+    Data,
     Dialect,
     Operation,
     ParametrizedAttribute,
@@ -31,9 +33,35 @@ from xdsl.irdl import (
     result_def,
     var_operand_def,
 )
-from xdsl.parser import Parser
+from xdsl.parser import AttrParser, Parser
 from xdsl.printer import Printer
 from xdsl.traits import SymbolOpInterface
+
+
+class EffectsEnum(Enum):
+    NONE = "none"
+    FULL = "full"
+
+
+class EffectsAttr(Data[EffectsEnum]):
+    """
+    An Attribute specifying if the marked operation has any effects on accelerator states.
+    """
+
+    name = "accfg.effects"
+
+    @classmethod
+    def parse_parameter(cls, parser: AttrParser) -> EffectsEnum:
+        with parser.in_angle_brackets():
+            for field in EffectsEnum:
+                if parser.parse_optional_keyword(field.value):
+                    return field
+            valid_vals = ", ".join(field.value for field in EffectsEnum)
+            parser.raise_error(f"Unknown keyword, expected one of: {valid_vals}")
+
+    def print_parameter(self, printer: Printer) -> None:
+        with printer.in_angle_brackets():
+            printer.print(self.data.value)
 
 
 @irdl_attr_definition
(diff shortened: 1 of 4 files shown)


AntonLydike:
How difficult is it to update to a version past 3.10?

AntonLydike:
image

AntonLydike:
Yes, got a workaround working now!

AntonLydike:
But we'll run into this eventually as StringEnum is used all over the place (also soon in llvm and arith dialect if I'm not mistaken)

jorendumoulin:
In xdsl or mlir?

AntonLydike:
> In xdsl or mlir?

Once xdsl targets a newer MLIR we need fastmath flags and such, and for that (as far as I am aware) the current goal is to use StrEnum or derivatives.

compiler/inference/helpers.py
```diff
@@ -1,11 +1,44 @@
...
 
 from compiler.dialects import accfg
 
 
+def has_accfg_effects(op: Operation) -> bool:
+    """
+    Checks if an operation effects state managed by the accfg dialect, according to the following criteria:
+
+    - If provided, an attribute named `accfg_effects` of either of the two types will overwrite inference.
+      These attributes will need to be added by the provider of the IR, as only they know which functions may affect
```
JosseVanDelm, line 14: Can we as "providers of the IR" just add attributes to any operation? Will this not introduce parsing errors?
E.g. A `memref.subview` operation with a `accfg_effect` attribute is not a valid `memref.subview` operation anymore, is it?

compiler/dialects/accfg.py
```diff
@@ -31,11 +33,37 @@
...
 )
-from xdsl.parser import Parser
+from xdsl.parser import AttrParser, Parser
 from xdsl.printer import Printer
 from xdsl.traits import SymbolOpInterface
 
 
+class EffectsEnum(Enum):
+    NONE = "none"
+    FULL = "full"
```
JosseVanDelm, line 43:
```suggestion
    """
    Specify to explicitly say:
    * NONE: There are no accfg side-effects to this operation
    * FULL: There are accfg side-effects to this operation
    """
    NONE = "none"
    FULL = "full"
```

JosseVanDelm (commented):
I'm not fully sure this can work?

JosseVanDelm:
> It is mildly to significantly annoying, mostly because ubuntu 22.04 ships with 3.10 and more recent versions of ubuntu have no good llvm support yet, and installing other versions of python on ubuntu is not fun. We're going to move to ubuntu 24 as soon as the llvm issues are resolved, which ships with python 3.12

Another reason not to move to 3.12 is that tensorflow currently does not do 3.12, and we use it to see if our input IR is still parseable by xdsl in the MLPerf Tiny tests.

compiler/inference/helpers.py
AntonLydike: An operation can have arbitrary attributes, they are meant to be used like this. Properties are "fixed", as in there is a fixed set of them per operation, but you can set arbitrary attributes on any operation.

AntonLydike:
> > It is mildly to significantly annoying, mostly because ubuntu 22.04 ships with 3.10 and more recent versions of ubuntu have no good llvm support yet, and installing other versions of python on ubuntu is not fun. We're going to move to ubuntu 24 as soon as the llvm issues are resolved, which ships with python 3.12
> 
> Another reason not to move to 3.12 is that tensorflow currently does not do 3.12, and we use it to see if our input IR is still parseable by xdsl in the MLPerf Tiny tests.

Big ooof. Well, luckily for now we don't need to update^^

> tf nightly does support 3.12 it seems. But I'm not sure we want that as a dependency 😄

well, shouldn't be long until it's released then?

AntonLydike:
@JosseVanDelm can I get a :+1: if this is gtg?

JosseVanDelm (approved):
Can you add a test for this please?

PR #169, opened 2024-07-01
support dynamic lcbs

jorendumoulin:
When looking for a largest contiguous block, we were too conservative when bumping into dynamic dimensions. This results in a copy from a `memref<?xi32, strided[1]>` to be copied in chunks of single elements, even though the data is still contiguous.

This PR should hopefully fix this.

### DIFF
--- a/compiler/ir/tsl/tiled_strided_layout.py
+++ b/compiler/ir/tsl/tiled_strided_layout.py
@@ -148,6 +148,7 @@ class TiledStridedLayout:
             contiguous block.
 
         """
+        self_strides = [x for x in self]
         result: list[Stride] = []
 
         # provide default result of single element common
@@ -161,7 +162,7 @@ class TiledStridedLayout:
             next_stride = next(
                 (
                     (dim, depth, stride_self)
-                    for dim, depth, stride_self in self
+                    for dim, depth, stride_self in self_strides
                     if stride_self.step == current_stride
                 ),
                 None,
@@ -170,17 +171,18 @@ class TiledStridedLayout:
             # check if contiguous block is found
             if next_stride is None:
                 return result or default_result
-            # return if dynamic
-            if next_stride[2].is_dynamic():
-                return result or default_result
             else:
                 dim, depth, stride_self = next_stride
+                self_strides.remove(next_stride)
 
             # check if contiguous block is common with other layout
             stride_other = other.get_stride(dim, depth)
             if stride_self == stride_other:
                 result.append(stride_self)
-                current_stride = stride_self.step * stride_self.bound
+                if stride_self.step is None or stride_self.bound is None:
+                    current_stride = None
+                else:
+                    current_stride = stride_self.step * stride_self.bound
 
             else:
                 return result or default_result
--- a/compiler/transforms/snax_copy_to_dma.py
+++ b/compiler/transforms/snax_copy_to_dma.py
@@ -4,6 +4,7 @@ from xdsl.dialects.builtin import (
     IndexType,
     IntAttr,
     IntegerType,
+    MemRefType,
     NoneAttr,
     StridedLayoutAttr,
 )
@@ -12,9 +13,8 @@ from xdsl.dialects.memref import (
     Dim,
     ExtractAlignedPointerAsIndexOp,
     ExtractStridedMetaDataOp,
-    MemRefType,
 )
-from xdsl.ir import Block, MLContext, Region
+from xdsl.ir import Block, MLContext, Region, SSAValue
 from xdsl.passes import ModulePass
 from xdsl.pattern_rewriter import (
     PatternRewriter,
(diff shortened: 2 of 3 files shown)


compiler/ir/tsl/tiled_strided_layout.py
```diff
@@ -170,17 +171,18 @@ def largest_common_contiguous_block(
             # check if contiguous block is found
             if next_stride is None:
                 return result or default_result
-            # return if dynamic
-            if next_stride[2].is_dynamic():
-                return result or default_result
             else:
```
jorendumoulin, line 174: This is the main change of the PR, to let this continue if it encounters something dynamic.

compiler/transforms/snax_copy_to_dma.py
```diff
@@ -28,6 +28,42 @@
 from compiler.ir.tsl import TiledStridedLayout
 
 
+def get_total_size_op(source: SSAValue):
```
jorendumoulin, line 31: This is code from `MatchSimpleCopy` factored out as we also need it if we emit a 1D transfer in `TransformCopy`

compiler/transforms/snax_copy_to_dma.py
```diff
@@ -336,17 +345,27 @@ def match_and_rewrite(self, op: CopyOp, rewriter: PatternRewriter):
...
         if len(remaining_strides) == 0:
-            # is actually 1d dma transfer, fake 2d transfer for experiments
-            one_op = arith.Constant.from_int_and_width(1, IndexType())
-            ops_to_insert.append(one_op)
-            dma_loop = {
-                "step_src_op": one_op,
-                "step_dst_op": one_op,
-                "bound_op": one_op,
-            }
+            # is actually 1d dma transfer, calculate size of transfer:
```
jorendumoulin, line 348: lol, this is fixed now

tests/filecheck/transforms/copy_to_dma.mlir
```diff
@@ -55,6 +55,54 @@
 
 // -----
 
+"builtin.module"() ({
+  "func.func"() <{"sym_name" = "transform_copy", "function_type" = (memref<?x?xi32, strided<[?, 1]>, "L0">, memref<?x?xi32, strided<[?, 1]>, "L1">) -> (), "sym_visibility" = "public"}> ({
+  ^0(%arg0 : memref<?x?xi32, strided<[?, 1]>, "L0">, %arg1 : memref<?x?xi32, strided<[?, 1]>, "L1">):
+    "memref.copy"(%arg0, %arg1) : (memref<?x?xi32, strided<[?, 1]>, "L0">, memref<?x?xi32, strided<[?, 1]>, "L1">) -> ()
+    "func.return"() : () -> ()
+  }) : () -> ()
+}) : () -> ()
```
JosseVanDelm, line 64: Can you add an extra test for strided memrefs with an offset please?

compiler/ir/tsl/tiled_strided_layout.py
```diff
@@ -161,7 +162,7 @@ def largest_common_contiguous_block(
             next_stride = next(
                 (
                     (dim, depth, stride_self)
-                    for dim, depth, stride_self in self
+                    for dim, depth, stride_self in self_strides
```
JosseVanDelm, line 165: This change along with `[x for x in self]` seems unnecessary?

compiler/transforms/snax_copy_to_dma.py
```diff
@@ -28,6 +28,42 @@
...
+    for dim in range(source.type.get_num_dims()):
+        const_op = Constant.from_int_and_width(dim, IndexType())
+        ops_to_insert.append(const_op)
+        dim_op = Dim.from_source_and_index(source, const_op.result)
+        ops_to_insert.append(dim_op)
+        if total_size_op is None:
+            total_size_op = dim_op
+        else:
+            total_size_op = Muli(total_size_op.result, dim_op.result, IndexType())
+            ops_to_insert.append(total_size_op)
```
JosseVanDelm, line 47: Can you move this to a utility function (maybe separate PR) ? It is used in a lot of places

JosseVanDelm (changes requested):
I'm going to put my foot down for adding offset tests this time :D

compiler/transforms/snax_copy_to_dma.py
jorendumoulin: Ok, but where is it used elsewhere?

compiler/ir/tsl/tiled_strided_layout.py
jorendumoulin: It is necessary because we now also search for dynamic strides.

In the static scenario, we search for a stride (for example 4), and then increment the stride (to 16).
In the next step, we search for stride 16, and then increment again if it is found.

However in the dynamic case, the stride is None, and remains None. So now, we must make sure we do find the same stride twice. Therefore, a hit is removed from this list `self_strides`, as you can see on line 176

jorendumoulin:
> I'm going to put my foot down for adding offset tests this time :D

fair enough 😄

compiler/ir/tsl/tiled_strided_layout.py
JosseVanDelm: Oh oops actually I was just a bit confused. You are obviously using the `__iter__` method here :laughing: nevermind the previous comment

compiler/transforms/snax_copy_to_dma.py
JosseVanDelm: oh wait I'm probably overreacting :dango:

jorendumoulin:
I added a test for the dynamic offsets in strided attributes! Although this filecheck is getting a bit out of nad.
I think in a following PR I'll try and add some cse and cannonicalization if possible to clean this up a bit

tests/filecheck/transforms/copy_to_dma.mlir
JosseVanDelm, line 1: I guess it doesn't matter that we are not using L1 and L3 for this test here

tests/filecheck/transforms/copy_to_dma.mlir
```diff
@@ -55,6 +55,95 @@
...
+//CHECK-NEXT:     %6 = "arith.constant"() <{"value" = 1 : index}> : () -> index
+//CHECK-NEXT:     %7 = "arith.divui"(%3, %6) : (index, index) -> index
+//CHECK-NEXT:     %8 = "arith.constant"() <{"value" = 1 : index}> : () -> index
+//CHECK-NEXT:     %9 = "arith.divui"(%5, %8) : (index, index) -> index
+//CHECK-NEXT:     %10 = "arith.constant"() <{"value" = 4 : index}> : () -> index
+//CHECK-NEXT:     %11 = "arith.constant"() <{"value" = 4 : index}> : () -> index
+//CHECK-NEXT:     %12 = "arith.muli"(%9, %10) : (index, index) -> index
+//CHECK-NEXT:     %13 = "arith.constant"() <{"value" = 4 : index}> : () -> index
+//CHECK-NEXT:     %14 = "arith.constant"() <{"value" = 4 : index}> : () -> index
+//CHECK-NEXT:     %15 = "arith.muli"(%9, %13) : (index, index) -> index
```
JosseVanDelm, line 125: if the algorithm detects that it is contiguous, then we can maybe elide emitting this?
I know this will get dce'd, but maybe we don't even have to insert it in the first place?

JosseVanDelm (approved):
Thank you :heart_hands:

tests/filecheck/transforms/copy_to_dma.mlir
jorendumoulin: I'll keep this in mind, but I think it is quite difficult to do this. The code is already complicated enough, and if dce works properly it is not worth it to complicate this further imo

JosseVanDelm: Sure! I don't think it's a priority.
Just because you said:
> Although this filecheck is getting a bit out of nad.

PR #170, opened 2024-07-08
bump tensorflow, snax, xdsl

jorendumoulin:
Tensorflow 2.16 should have Python 2.12 support!
This requires a new version of mlir
That requires a future version of xdsl

### DIFF
--- a/.github/workflows/build-run-kernel.yml
+++ b/.github/workflows/build-run-kernel.yml
@@ -10,7 +10,7 @@ jobs:
   build-and-run-kernels:
     runs-on: ubuntu-latest
     container: 
-      image: ghcr.io/kuleuven-micas/snax:v0.1.4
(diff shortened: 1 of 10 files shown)


jorendumoulin:
getting some errors since tensorflow uses a more recent version of mlir that includes this change: [lllvm/llvm-project#71084 ](https://github.com/llvm/llvm-project/pull/71084)

image
_Not Done_ 😢 

I am upgrading llvm at kuleuven-micas/snax_cluster#176

jorendumoulin:
All of the remaining errors at this point seem to be related to the `IntegerOvewrflowFlagsAttr` not yet implemented in xdsl. @AntonLydike are these the flags you were referring to that were going to use the `StrEnum` stuff?

AntonLydike:
I think so, yes, but I'm not 100% sure. I opened a new discussion on Zulip that we update to a newer MLIR again

PR #172, opened 2024-07-09
switch to filecheck-ng

AntonLydike:
Fixes #171

PR #173, opened 2024-07-09
Add reset op and accfg-insert-resets

AntonLydike:
This adds a reset op and a new pass called `accfg-insert-resets`, which adds an `accfg.reset` operation onto every dangling state.

Stacked onto #172


compiler/transforms/accfg_insert_resets.py
```diff
@@ -0,0 +1,113 @@
...
+                if not isinstance(val.type, val_type):
+                    continue
+                if val in seen:
+                    continue
+                seen.add(val)
+                fun(self, val, rewriter)
+
+        return match_and_rewrite
+
+    return wrapper
```
AntonLydike, line 36: I want to eventually move this into upstream xDSL

tests/filecheck/transforms/accfg-insert-resets.mlir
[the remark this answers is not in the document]
JosseVanDelm: Oh this one is in the other test as well?
Shouldn't  we also test this without overlap?

compiler/dialects/accfg.py
```diff
@@ -430,12 +430,29 @@ def launch_field_items(self) -> Iterable[tuple[str, IntegerAttr]]:
...
+    in_state = operand_def(StateType)
+
+    assembly_format = "$in_state attr-dict `:` type($in_state)"
+
+    def __init__(self, in_state: Operation | SSAValue):
+        super().__init__(operands=[in_state])
+
+    def get_modified_fields(self) -> set[str]:
+        # TODO: implement this part
+        return set()
```
JosseVanDelm, line 446: why is this here?

compiler/transforms/accfg_insert_resets.py
```diff
@@ -0,0 +1,126 @@
...
+        inserts[block] = InsertCandidate(idx, use.operation)
+
+    # return all insertion candidates that we have
+    yield from (InsertPoint.after(cd.op) for cd in inserts.values())
+
+
+class InsertResetsForDanglingStatesPattern(RewritePattern):
+    """
+    This inspects all SSA values and inserts a `reset` operation if the value has no uses outside of `launch` and
+    `reset` operations.
```
JosseVanDelm, line 97: why also on reset operations? Wouldn't  this reset twice?

compiler/transforms/accfg_insert_resets.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+
+    @ssa_val_rewrite_pattern(accfg.StateType)
+    def match_and_rewrite(self, val: SSAValue, rewriter: PatternRewriter, /):
+        # abort if:
+        #  - there are uses of the op
+        #  - and all of them are setups/reset ops
+        # this may not be correct, let's see...
```
JosseVanDelm, line 105: I'm not sure I understand this

JosseVanDelm (commented):
I have a few questions

tests/filecheck/transforms/accfg-insert-resets.mlir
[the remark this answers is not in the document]
AntonLydike: Hmmm, probably yes. Will add!

tests/filecheck/transforms/accfg-insert-resets.mlir
```diff
@@ -0,0 +1,111 @@
...
+    %t = "accfg.launch"(%state) <{"param_names" = [], "accelerator" = "acc"}> : (!accfg.state<"acc">) -> !accfg.token<"acc">
+    "accfg.await"(%t) : (!accfg.token<"acc">) -> ()
+    return
+}
+
+// CHECK-LABEL: @with_uses
+// CHECK-NEXT: [[state:%\S+]] = accfg.setup "acc" to ("i" = %i : i32) : !accfg.state<"acc">
+// CHECK-NEXT: [[token:%\S+]] = "accfg.launch"(%state)
+// CHECK-NEXT: accfg.reset [[state]] : !accfg.state<"acc">
+// CHECK-NEXT: "accfg.await"([[token]])
```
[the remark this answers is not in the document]
AntonLydike: The `CHECK` lines don't need to match the whole line

compiler/dialects/accfg.py
AntonLydike: good question! I originally wanted to have the lowering of the reset op in this PR as well, will remove!

compiler/transforms/accfg_insert_resets.py
AntonLydike: Yes, correct, this docstring seems to be wrong. I do need to figure out exactly when the resets are inserted!

compiler/transforms/accfg_insert_resets.py
AntonLydike: Me neither, will look at this tomorrow morning or so again when I'm fresh in the brain I think. So far it *seems* to work, according to the tests, but I'm not convinced the check is actually correct...

compiler/transforms/accfg_insert_resets.py
```diff
@@ -0,0 +1,126 @@
...
+    PatternRewriteWalker,
+    RewritePattern,
+)
+from xdsl.rewriter import InsertPoint
+from xdsl.traits import IsTerminator
+
+from compiler.dialects import accfg
+
+
+def ssa_val_rewrite_pattern(val_type: type[Attribute]):
```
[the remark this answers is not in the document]
AntonLydike: Done!

compiler/transforms/accfg_insert_resets.py
AntonLydike: I cleaned it up, I have a much better feeling about this now than I had about the other code. Added more tests to reassure myself again.

compiler/accelerators/accelerator.py
```diff
@@ -74,10 +74,3 @@ def lower_acc_setup(
         accelerator configuration.
         """
         raise NotImplementedError()
-
-    @staticmethod
-    @abstractmethod
-    def lower_reset(
-        rest_op: accfg.ResetOp, acc_op: accfg.AcceleratorOp
-    ) -> Sequence[Operation]:
```
JosseVanDelm: This is moving to a next PR, right?

tests/filecheck/transforms/accfg-insert-resets.mlir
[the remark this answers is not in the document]
JosseVanDelm: I mainly thought it was weird that you already reset the accelerator's registers before the await.
I think we should be careful with this, since there could (and should) be hardware support for clearing all the registers (also the shadow registers) upon reset. That would make this code invalid if compiled. We should be very clear on semantics here, is all that I'm saying.
it is valid if the shadow registers are kept, but might lead to weird results if the reset affects them or other accelerator parameters. Maybe `reset` is too broad as a term?

compiler/transforms/accfg_insert_resets.py
```diff
@@ -1,35 +1,49 @@
...
+    Callable[[_RewritePatternT, Operation, PatternRewriter], None],
+]:
+    """
+    Expresses a rewrite pattern that acts on SSA Values instead of operations.
+
+    Pass `val_type` and only match on SSA Values of the specified type.
+    """
+
+    def wrapper(
+        old_match_and_rewrite: Callable[
```
JosseVanDelm, line 31:
```suggestion
        wrapped_match_and_rewrite: Callable[
```

compiler/transforms/accfg_insert_resets.py
```diff
@@ -99,12 +113,10 @@ class InsertResetsForDanglingStatesPattern(RewritePattern):
...
-        # abort if:
-        #  - there are uses of the op
-        #  - and all of them are setups/reset ops
-        # this may not be correct, let's see...
-        if val.uses and all(
-            isinstance(use.operation, accfg.SetupOp | accfg.ResetOp) for use in val.uses
+        # abort if
+        # any use is a reset or another setup op
+        if any(
+            isinstance(use.operation, accfg.ResetOp | accfg.SetupOp) for use in val.uses
```
JosseVanDelm, line 119: this makes way more sense! Thanks!

compiler/accelerators/accelerator.py
AntonLydike: yes!

tests/filecheck/transforms/accfg-insert-resets.mlir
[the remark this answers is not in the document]
AntonLydike: Added a flag `reset-after-await` to pass, will need some more fine tuning, but already most of the way there I think

compiler/transforms/accfg_insert_resets.py
```diff
@@ -0,0 +1,126 @@
...
+    # if no uses are found, just return the first position where the value is live:
+    if not val.uses:
+        if isinstance(val.owner, Operation):
+            yield InsertPoint.after(val.owner)
+        else:
+            yield InsertPoint.at_start(val.owner)
+        return
+
+    # remember insertion candidates and their positions
+    # we need to keep track of the last use of the value in each block it's used in.
```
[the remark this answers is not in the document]
AntonLydike: So that we know where it's dangling, basically.

compiler/inference/dataflow.py
```diff
@@ -0,0 +1,107 @@
...
+def get_insertion_points_where_val_dangles(
+    vals: Sequence[SSAValue],
+) -> Iterable[InsertPoint]:
+    """
+    Return all insertion points for positions where a value starts dangling.
+
+    If val is used by a terminator, it's not assumed dangling in that block.
+
+    If multiple vals are given, return the earliest point in each block where
+    all vals used in that block are dangling.
```
JosseVanDelm, line 60: Why would we need to have multiple vals here? Is this because we both wait for the token and the state?
~~I find this quite difficult to read, would it make more sense to not make this a generator instead?~~
I should just read more generators haha

tests/filecheck/transforms/accfg-insert-resets.mlir
```diff
@@ -0,0 +1,133 @@
...
+        %0 = "accfg.launch"(%state) <{"param_names" = [], "accelerator" = "acc"}> : (!accfg.state<"acc">) -> !accfg.token<"acc">
+        "accfg.await"(%0) : (!accfg.token<"acc">) -> ()
+
+        yield
+    }, {
+        %1 = "accfg.launch"(%state) <{"param_names" = [], "accelerator" = "acc"}> : (!accfg.state<"acc">) -> !accfg.token<"acc">
+        "accfg.await"(%1) : (!accfg.token<"acc">) -> ()
+
+        yield
+    }) : (i1) -> ()
```
JosseVanDelm, line 69: Does it also work if the accelerator is not launched and awaited in one of the states?

tests/filecheck/transforms/accfg-insert-resets.mlir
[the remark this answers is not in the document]
JosseVanDelm: Is it possible you haven't added this yet?

compiler/transforms/accfg_insert_resets.py
```diff
@@ -0,0 +1,112 @@
...
+
+        # if reset_after_await is given, reset after the tokens of the launch ops are no longer dangling
+        # (i.e. have been awaited)
+        if self.reset_after_await:
+            vals = []
+            for use in uses:
+                if isinstance(use.operation, accfg.LaunchOp):
+                    vals.append(use.operation.token)
+            if not vals:
+                vals.append(val)
```
JosseVanDelm, line 87: what does this do?

compiler/inference/dataflow.py
```diff
@@ -0,0 +1,107 @@
...
+        - yield %x
+        - arith.add %a, ...
+    """
+    for use in val.uses:
+        # return each use
+        yield use
+        # if it's an scf.for, recurse on block arg
+        if isinstance(use.operation, scf.For):
+            for_op = use.operation
+            yield from uses_through_controlflow(for_op.body.block.args[use.index - 2])
```
JosseVanDelm, line 43: I don't understand why we need the -2?
Don't we want something like this?
```suggestion
            yield from uses_through_controlflow(for_op.body.block.args[use.index - 2:])
```

compiler/transforms/accfg_insert_resets.py
```diff
@@ -0,0 +1,112 @@
...
+        ):
+            return
+
+        # if reset_after_await is given, reset after the tokens of the launch ops are no longer dangling
+        # (i.e. have been awaited)
+        if self.reset_after_await:
+            vals = []
+            for use in uses:
+                if isinstance(use.operation, accfg.LaunchOp):
+                    vals.append(use.operation.token)
```
JosseVanDelm, line 85: I think the setup state itself should also be appended here, no?

JosseVanDelm (approved):
I find this quite difficult to review :( so feel free if my comments don't make sense
I have a few more comments, the results look good though!

tests/filecheck/transforms/accfg-insert-resets.mlir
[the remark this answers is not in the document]
AntonLydike: Oh damn, you are right!

tests/filecheck/transforms/accfg-insert-resets.mlir
AntonLydike: Interesting question, will investigate tomorrow morning

compiler/inference/dataflow.py
AntonLydike: Good point, yes. I should add more comments here.

The gist is that we want to get the block arg which represents the input arg, and -2 does that mapping. (`index - 3` is the idx of the loop carried variables, add one back because the first block arg is the loop index, and we end up at `index - 2`)

compiler/transforms/accfg_insert_resets.py
AntonLydike: Hmmmmmm.

it doesn't make sense in my mind. But making that change breaks no tests, and I can't think of a case where it would break.

I'll give it some thought over lunch

tests/filecheck/transforms/accfg-insert-resets.mlir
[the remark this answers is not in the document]
AntonLydike: Added!

compiler/inference/dataflow.py
JosseVanDelm: But so now it only works if there is just one block arg, right?

compiler/transforms/accfg_insert_resets.py
JosseVanDelm: I think it might break this test that I'm proposing:
https://github.com/KULeuven-MICAS/snax-mlir/pull/173#discussion_r1681233566

JosseVanDelm: ping @AntonLydike, can you take a look at this?

JosseVanDelm:
ping @AntonLydike !

compiler/inference/dataflow.py
AntonLydike: I improved/fixed a bunch of stuff, feel free to have a look again!

PR #174, opened 2024-07-09
Add tiled matmul example

JonasCrols:
Added support for snax-stream-gemm, used by adding a matmul that is tiled by using rectangular non-accumulating tiles.


runtime/Makefile.rules
```diff
@@ -125,6 +125,7 @@ MLIROPTFLAGS += --llvm-request-c-wrappers
 MLIROPTFLAGS += --expand-strided-metadata
 MLIROPTFLAGS += --convert-index-to-llvm=index-bitwidth=32
 MLIROPTFLAGS += --convert-cf-to-llvm=index-bitwidth=32
+MLIROPTFLAGS += --lower-affine
```
JosseVanDelm, line 128: Why do we need this?

JosseVanDelm (commented):
Thanks for your work @JonasCrols, could you still change the loops to be non-inter-accumulating and also make your PR description a bit more descriptive?

Also, why do we suddenly need `lower-affine?` Thanks!

runtime/Makefile.rules
jorendumoulin: If i remember correctly, there were some leftover affine.apply ops which were throwing errors down the line. I'm not sure where they appear and why the previous --lower-affine does not catch them, but adding this pass resolves the errors.

JosseVanDelm (approved):
Please change compilation pass order and remove non-functional example :heart:

tests/filecheck/transforms/set-memory-layout.mlir
```diff
@@ -5,7 +5,7 @@
   ^0(%arg0 : memref<?x128xi8, 1 : i32>, %arg1 : memref<128x128xi8, 1 : i32>, %arg2 : memref<?x128xi32, 1 : i32>):
     %0 = "arith.constant"() <{"value" = 1 : i32}> : () -> i32
     %1 = "arith.constant"() <{"value" = 1 : i32}> : () -> i32
-    "linalg.generic"(%arg0, %arg1, %0, %1, %arg2) <{"indexing_maps" = [affine_map<(d0, d1, d2) -> (d0, d2)>, affine_map<(d0, d1, d2) -> (d2, d1)>, affine_map<(d0, d1, d2) -> ()>, affine_map<(d0, d1, d2) -> ()>, affine_map<(d0, d1, d2) -> (d0, d1)>], "iterator_types" = [#linalg.iterator_type<parallel>, #linalg.iterator_type<parallel>, #linalg.iterator_type<reduction>], "operandSegmentSizes" = array<i32: 4, 1>, "library_call" = "snax_gemm"}> ({
```
JosseVanDelm, line 8: This change seems unnecessary

tests/filecheck/transforms/dispatch_regions.mlir
```diff
@@ -113,4 +113,43 @@
...
+    } ins(%0, %1 : memref<64xi32>, memref<64xi32>) outs(%2 : memref<64xi32>) {
+    ^bb0(%arg0: i32, %arg1: i32, %arg2: i32):
+      %3 = arith.muli %arg0, %arg1 : i32
+      linalg.yield %3 : i32
+  }
+    "func.return"() : () -> ()
+
+  }) : () -> ()
+}) : () -> ()
+
```
JosseVanDelm, line 135: how is this testing something different from the above test?

kernels/simple_matmul/main.c
```diff
@@ -21,8 +21,9 @@
  * /target/snitch_cluster/sw/snax/mac/include"
  *
  * */
-#include "snax-gemm-lib.h"
```
JosseVanDelm, line 24: Can we not include these routines now, instead of copy-pasting them?

tests/filecheck/transforms/dispatch_kernels.mlir
```diff
@@ -44,7 +44,7 @@
 //CHECK-NEXT:   ^0(%arg0 : memref<?x128xi8>, %arg1 : memref<128x128xi8>, %arg2 : memref<?x128xi32>):
 //CHECK-NEXT:     %0 = "arith.constant"() <{"value" = 1 : i32}> : () -> i32
 //CHECK-NEXT:     %1 = "arith.constant"() <{"value" = 1 : i32}> : () -> i32
-//CHECK-NEXT:     "linalg.generic"(%arg0, %arg1, %0, %1, %arg2) <{"indexing_maps" = [affine_map<(d0, d1, d2) -> (d0, d2)>, affine_map<(d0, d1, d2) -> (d2, d1)>, affine_map<(d0, d1, d2) -> ()>, affine_map<(d0, d1, d2) -> ()>, affine_map<(d0, d1, d2) -> (d0, d1)>], "iterator_types" = [#linalg.iterator_type<parallel>, #linalg.iterator_type<parallel>, #linalg.iterator_type<reduction>], "operandSegmentSizes" = array<i32: 4, 1>, "library_call" = "snax_gemm"}> ({
+//CHECK-NEXT:     "linalg.generic"(%arg0, %arg1, %0, %1, %arg2) <{"indexing_maps" = [affine_map<(d0, d1, d2) -> (d0, d2)>, affine_map<(d0, d1, d2) -> (d2, d1)>, affine_map<(d0, d1, d2) -> ()>, affine_map<(d0, d1, d2) -> ()>, affine_map<(d0, d1, d2) -> (d0, d1)>], "iterator_types" = [#linalg.iterator_type<parallel>, #linalg.iterator_type<parallel>, #linalg.iterator_type<reduction>], "operandSegmentSizes" = array<i32: 4, 1>, "library_call" = "snax_qgemm"}> ({
```
JosseVanDelm, line 47: Why is this changed?

kernels/half_tiled_matmul/main.c
```diff
@@ -0,0 +1,209 @@
...
+    write_csr(987, 0);
+    write_csr(987, 0);
+    write_csr(981, 0);
+
+    // // implement some artificial delay
+    // for (int i = 0; i < 1000; i++) {{
+    //     asm volatile("nop");
+    // }}
+  }
+}
```
JosseVanDelm, line 125: Same comment here, can we please include these instead of putting them here?

.github/workflows/build-run-kernel.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   build-and-run-kernels:
     runs-on: ubuntu-latest
     container: 
-      image: ghcr.io/kuleuven-micas/snax:v0.1.4
+      image: ghcr.io/kuleuven-micas/snax:v0.1.6
```
JosseVanDelm, line 13: I'd prefer updating all CI routines in one go to use the newest container.
Can you also update the other CI jobs please? Thanks!

kernels/half_tiled_matmul/half_tiled_matmul.mlir
```diff
@@ -0,0 +1,23 @@
...
+      %tiled_D = memref.subview %arg2[%iv, %jv][%short_tile_size, %short_tile_size][%c1, %c1] : memref<16x16xi32> to memref<?x?xi32, strided<[16, 1], offset: ?>>
+      
+      // Perform the quantized matrix multiplication on the tiles
+      linalg.quantized_matmul ins(%tiled_A, %tiled_B, %c0_i32, %c0_i32 : memref<?x?xi8, strided<[16, 1], offset: ?>>, memref<?x?xi8, strided<[1, 16], offset: ?>>, i32, i32) outs(%tiled_D : memref<?x?xi32, strided<[16, 1], offset: ?>>)
+      
+    }
+  }
+  
+  return
+}
```
JosseVanDelm, line 23: add newline

JosseVanDelm (changes requested):
Hey this is nicely coming along!
Hi, sorry but I'm noticing a lot of code copy pasting of which I'm not a really big fan.
I'm also not sure why you changed the library call to `snax_qgemm`, but  feel free to go ahead if this is what you prefer.
I'm also not entirely sure the test you put here tests the previous error?
There are also still a few left-over suggestions from Joren.

tests/filecheck/transforms/dispatch_regions.mlir
JonasCrols: the Dispatch_regions pass fails only when Linalg.generic is used in this way. when it is used in the other way (with quotation marks), the pass always performs correctly. we now test both cases

tests/filecheck/transforms/dispatch_kernels.mlir
JonasCrols: because the new gemm-call is called snax_qgemm now. is there a way this doesn't need to be changed?

kernels/simple_matmul/main.c
JonasCrols: what exactly would you like included in a separate routine? most of this comes straight out of the template

tests/filecheck/transforms/dispatch_regions.mlir
AntonLydike: That shouldn't be though, the quotation marks just mark generic vs. custom syntax of how the operation is printed. They should represent the exact same operation though...

kernels/simple_matmul/main.c
```diff
@@ -104,4 +104,4 @@ int main() {
       nerr += 1;
   }
   return nerr;
-}
+}
```
jorendumoulin, line 107: add newline

runtime/snax-streamer-gemm.rules
```diff
@@ -0,0 +1,5 @@
+# Specific settings for snax-mac RTL
```
jorendumoulin, line 1:
```suggestion
# Specific settings for snax-streamer-gemm RTL
```

jorendumoulin (approved):
Perfect, thank you! 
Jus two last final nits!

kernels/streamer_matmul/Makefile
```diff
@@ -0,0 +1,40 @@
...
+TESTS =
+TESTS += half_tiled_matmul.x
+TESTS += matmul.x
+
+CFLAGS += -std=gnu11
+CFLAGS += -Wall -Wextra
+
+# Override snax-opt rules to avoid linalg-to-library-call pass
+SNAXOPTACCFLAGS = -p dispatch-kernels,set-memory-space,set-memory-layout,realize-memref-casts,insert-sync-barrier,dispatch-regions,insert-accfg-op{accelerator=snax_gemm},convert-linalg-to-accfg,convert-accfg-to-csr,snax-copy-to-dma,memref-to-snax,snax-to-func,clear-memory-space
+
```
JosseVanDelm, line 17: I think this is still running on CPU, can you remove the `*accfg*` passes and insert `convert-linalg-to-library-call` pass?

JosseVanDelm (changes requested):
I think it is not using the accelerator yet

kernels/streamer_matmul/Makefile
```diff
@@ -0,0 +1,40 @@
...
+%.acc_dialect.snax-opt.mlir: %.preprocfinal.mlir
+	$(SNAXOPT) $(SNAXOPTACCFLAGS) -o $@ $<
+
```
JosseVanDelm, line 20: Remove these lines please

JosseVanDelm (approved):
Okay gotem

kernels/streamer_matmul/Makefile
JosseVanDelm: Turns out these passes are not used anyways :D

PR #175, opened 2024-07-10
add snax-stream dialect

jorendumoulin:
This PR adds the snax_stream dialect.
In many ways, it is a copy of the snitch_stream dialect, with the following changes to adapt it to SNAX Streamers:

- operate on indeces instead of riscv registers
- use a stride pattern with upper bounds, temporal strides, spatial strides

### DIFF
--- a/compiler/dialects/snax_stream.py
+++ b/compiler/dialects/snax_stream.py
@@ -22,6 +22,7 @@ from xdsl.irdl import (
 from xdsl.parser import AttrParser
 from xdsl.printer import Printer
 
+
 @irdl_attr_definition
 class StridePattern(ParametrizedAttribute):
     """
--- a/compiler/dialects/snax_stream.py
+++ b/compiler/dialects/snax_stream.py
@@ -1,7 +1,14 @@
 from collections.abc import Sequence
 
 from xdsl.dialects.builtin import ArrayAttr, IndexType, IntAttr
-from xdsl.ir import Attribute, Dialect, NoTerminator, ParametrizedAttribute, Region, SSAValue
+from xdsl.ir import (
+    Attribute,
+    Dialect,
+    NoTerminator,
+    ParametrizedAttribute,
+    Region,
+    SSAValue,
+)
 from xdsl.irdl import (
     AttrSizedOperandSegments,
     IRDLOperation,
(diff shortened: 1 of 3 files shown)


compiler/dialects/snax_stream.py
```diff
@@ -0,0 +1,132 @@
...
+class StreamingRegionOp(IRDLOperation):
+    name = "snax_stream.streaming_region"
+
+    # inputs and outputs are pointers to the base address of the data in memory
+    inputs = var_operand_def(IndexType)
+    outputs = var_operand_def(IndexType)
+
+    # streaming stride pattern
+    # there should be one stride pattern for every input/output
+    # the upper bounds of all stride patterns should be equal
```
JosseVanDelm, line 107: Can you add verifiers for this? Also, do all strides need to be positive?

compiler/dialects/snax_stream.py
```diff
@@ -0,0 +1,132 @@
...
+                IntAttr(i)
+                for i in parser.parse_comma_separated_list(
+                    parser.Delimiter.SQUARE, parser.parse_integer
+                )
+            )
+            return (ub, ts, ss)
+
+
+@irdl_op_definition
+class StreamingRegionOp(IRDLOperation):
```
JosseVanDelm, line 98: Please add a docstring :)

compiler/dialects/snax_stream.py
```diff
@@ -0,0 +1,132 @@
...
+class StridePattern(ParametrizedAttribute):
+    """
+    Attribute representing a strided streaming pattern in SNAX.
+    """
+
+    name = "snax_stream.stride_pattern"
+
+    upper_bounds: ParameterDef[ArrayAttr[IntAttr]]
+    temporal_strides: ParameterDef[ArrayAttr[IntAttr]]
+    spatial_strides: ParameterDef[ArrayAttr[IntAttr]]
```
JosseVanDelm, line 36: Please give a simple example here in the docstring to explain all parameters

JosseVanDelm (commented):
This looks super cool.
If at some point you implement the StructuredOpInterface we get free operator fusion :thinking: ?
Can you add  a bit more documentation (semantics of the op are not super clear now) and verification (and failing verification test) though please?

compiler/dialects/snax_stream.py
jorendumoulin: If it's all right I will do this later. The next step is to assign every snax_stream an accelerator which must have the streamer interface implemented. This interface has all the information how the streamers look like and gives us a much better information for verification of the stride pattern.

JosseVanDelm: Okay to do later, but not sure I agree with your verification plan.
There's a bit of verification you can already do without lowering to a specific accelerator.
If you insert an incorrect operation of this kind at some point in a pass, you would want the verifier to already report this before lowering. The things you describe in the comments above should be possible to check without any accelerator information, no?

jorendumoulin: Yes but this op can only be lowered to if there is an accelerator with streamers in the system. I think I'll add the operatator accfg definition as an operand of this op.

PR #176, opened 2024-07-12
Add snax streamer interface

jorendumoulin:
note: stacked on #175 

This PR includes the following:

- A representation of the Streamers in `accelerators/streamers/streamer.py`
- An attribute which holds as data a StreamerConfiguration: `StreamerConfigurationAttr`
- This attribute can be passed along with an `accfg.AcceleratorOp`.
- The `snax_stream.streaming_region_op` now comes with a verifier that does the following:
  - The region must be linked to an accelerator
  - The accelerator must implement a streamer interface, through the streamer configuration attribute
  - All of the dimensions of the stride pattern must mach the streamer interface implementation
- It also includes a `SNAXStreamer` base class, which can be inherited by other accelerators, this class will then handle anything streamer-related automatically, based on the supplied StreamerConfiguration. (This is implemented in future PRs, see #177 for a sneak peak)

### DIFF
--- a/compiler/dialects/snax_stream.py
+++ b/compiler/dialects/snax_stream.py
@@ -62,9 +62,13 @@ class StridePattern(ParametrizedAttribute):
             printer.print_string("ub = [")
             printer.print_list(self.upper_bounds, lambda attr: printer.print(attr.data))
             printer.print_string("], ts = [")
-            printer.print_list(self.temporal_strides, lambda attr: printer.print(attr.data))
+            printer.print_list(
+                self.temporal_strides, lambda attr: printer.print(attr.data)
+            )
             printer.print_string("], ss = [")
-            printer.print_list(self.spatial_strides, lambda attr: printer.print(attr.data))
+            printer.print_list(
+                self.spatial_strides, lambda attr: printer.print(attr.data)
+            )
             printer.print_string("]")
 
     @classmethod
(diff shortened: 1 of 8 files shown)


compiler/dialects/snax_stream.py
```diff
@@ -21,6 +22,11 @@
 )
 from xdsl.parser import AttrParser
 from xdsl.printer import Printer
+from xdsl.traits import SymbolTable
+
+from compiler.accelerators.streamers.streamers import StreamerConfiguration
```
JosseVanDelm, line 27: Are you planning to put a lot more stuff into streamers? Or can we do 
```suggestion
from compiler.accelerators.streamers import StreamerConfiguration
```

compiler/dialects/snax_stream.py
```diff
@@ -139,5 +155,47 @@ def __init__(
...
+
+        if not isinstance(acc_op, AcceleratorOp):
+            raise VerifyException("AcceleratorOp not found!")
+
+        streamer_interface = acc_op.get_attr_or_prop("streamer_config")
+        if not streamer_interface or not isinstance(
+            streamer_interface, StreamerConfigurationAttr
+        ):
+            raise VerifyException(
+                "Specified accelerator does not implement the streamer interface"
```
JosseVanDelm, line 177:
```suggestion
                "Specified accelerator does not contain a StreamerConfigAttr"
```

JosseVanDelm (approved):
Very cool!

compiler/dialects/snax_stream.py
jorendumoulin: we can import like this! auto-import fucked this one up for me. I put streamers in a folder for possible variations on the streamer (sparse-streamer, pointer-chasing-streamer, ...)

compiler/dialects/snax_stream.py
jorendumoulin: ok!

PR #177, opened 2024-07-15
add streamer default lowerings

jorendumoulin:


compiler/accelerators/snax.py
```diff
@@ -59,6 +61,10 @@ def lower_acc_setup(
         field_to_csr = dict(acc_op.field_items())
         ops: Sequence[Operation] = []
         for field, val in setup_op.iter_params():
+            if isinstance(val.type, builtin.IndexType):
+                val_to_i32 = arith.IndexCastOp(val, builtin.i32)
+                ops.append(val_to_i32)
+                val = val_to_i32.result
```
JosseVanDelm, line 67: I like this, but I'd prefer we split it in a more general (also parameterized) pass maybe? Can do this in a later PR.

compiler/accelerators/snax_alu.py
```diff
@@ -3,38 +3,51 @@
...
+    StreamerConfiguration,
+    StreamerType,
+)
+from compiler.dialects import accfg, snax_stream
 
+default_streamer = StreamerConfiguration(
+    [
+        Streamer(StreamerType.Reader, 1, 1),
+        Streamer(StreamerType.Reader, 1, 1),
+        Streamer(StreamerType.Reader, 1, 1),
```
JosseVanDelm, line 18:
```suggestion
        Streamer(StreamerType.Writer, 1, 1),
```

compiler/transforms/insert_accfg_op.py
```diff
@@ -31,9 +31,7 @@ def match_and_rewrite(self, op: builtin.ModuleOp, rewriter: PatternRewriter):
         assert t is not None
         acc_string = self.acc_op.name_prop.string_value()
         if t.lookup_symbol(op, acc_string) is not None:
-            raise RuntimeError(
-                f"Cannot insert op: @{acc_string}, already present in symbol table."
-            )
```
JosseVanDelm, line 36: Why are you removing this?

JosseVanDelm (commented):
Cool cool cool. I have a few comments :)

compiler/accelerators/snax_alu.py
jorendumoulin: aha yes

compiler/transforms/insert_accfg_op.py
jorendumoulin: In my test, I specify my own accfg op. I want to avoid having a different pipeline for every test, so I thought this would be preferenrtial behaviour: Insert the default accfg op, but if you want to override it with your own, then fine.

compiler/transforms/insert_sync_barrier.py
```diff
@@ -34,6 +34,10 @@ def match_and_rewrite(self, module: builtin.ModuleOp, rewriter: PatternRewriter)
                 # clear the list
                 ops_to_sync = []
 
+            if isinstance(op, snax.ClusterSyncOp):
+                # synchronisation ok, clear list
+                ops_to_sync = []
+
```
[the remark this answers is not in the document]
jorendumoulin: just to make my test work 😬 . The test is already mid-way in the pipeline, and thus has some tests applied 2x. Without this, it tries to add multiple synchronizations. Because some regions are already dispatched, this leads to some synchronization bugs.

I guess we should make sure applying a pass at the wrong time / applying it twice should never result in an error. This makes sure this is the case for this one.

compiler/transforms/insert_accfg_op.py
jorendumoulin: ok nvm i put it back

compiler/accelerators/snax.py
jorendumoulin: ok will do!

PR #178, opened 2024-07-15
allow unregistered accelerator ops

jorendumoulin:
I want to write tests with some arbitrary accelerator ops, but the system won't allow this unless they are registered. However, I don't want to register these dummy accelerator ops.

I therefore suggest we allow unregistered accelerator ops, and only throw runtime errors when we actually apply transformations on accfg ops that are not registered.

### DIFF
--- a/compiler/accelerators/registry.py
+++ b/compiler/accelerators/registry.py
@@ -26,7 +26,7 @@ class AcceleratorRegistry:
 
     def lookup_acc_info(
         self, acc_query: StringAttr, module: ModuleOp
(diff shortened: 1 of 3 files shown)


jorendumoulin:
If you don't like this change, I can alternatively split up the function `lookup_acc_info` into `get_acc_op` and `get_acc_type`. I only need `get_acc_op`.

I'm not sure what's best here

jorendumoulin:
ok nvm

jorendumoulin:
ok nvm

PR #180, opened 2024-07-17
fix dispatch regions

JonasCrols:
Fix bug where dispatch_regions would add if-statement inside the if-statements own region, thus creating a loop. This is solved using region_first option. a test was added


jorendumoulin (approved):
Perfect!

Thank you Jonas!

jorendumoulin:
Now smash ~~like and subscribe~~ squash and merge~!

JosseVanDelm (approved):
@JonasCrols Thank you!

PR #182, opened 2024-07-19
add stream-snaxify pass

jorendumoulin:
This pass converts a `memref.streaming_region_op` into a `snax_stream.streaming_region_op`. 
This is somewhat similar to the pass `convert_memref_stream_to_snitch_stream` in xdsl. I am refactoring some stuff over there to share some more code.

It also updates the lowering for the `snax_alu` accelerator to start from `memref_stream`.

### DIFF
--- a/compiler/dialects/snax_stream.py
+++ b/compiler/dialects/snax_stream.py
@@ -151,7 +151,7 @@ class StreamingRegionOp(IRDLOperation):
     ) -> None:
         if not isinstance(stride_patterns, ArrayAttr):
             stride_patterns = ArrayAttr(stride_patterns)
-        if not isinstance(accelerator, StringAttr):
+        if isinstance(accelerator, str):
             accelerator = StringAttr(accelerator)
         super().__init__(
             operands=[inputs, outputs],
--- a/compiler/dialects/snax_stream.py
+++ b/compiler/dialects/snax_stream.py
@@ -5,6 +5,7 @@ from xdsl.ir import (
     Attribute,
     Dialect,
     NoTerminator,
+    Operation,
     ParametrizedAttribute,
     Region,
     SSAValue,
@@ -127,7 +128,7 @@ class StreamingRegionOp(IRDLOperation):
 
     # streaming stride pattern
     # there should be one stride pattern for every input/output
-    stride_pattern = prop_def(ArrayAttr[StridePattern])
+    stride_patterns = prop_def(ArrayAttr[StridePattern])
 
     accelerator = prop_def(StringAttr)
     """
@@ -142,17 +143,20 @@ class StreamingRegionOp(IRDLOperation):
 
     def __init__(
         self,
-        inputs: Sequence[SSAValue],
-        outputs: Sequence[SSAValue],
+        inputs: Sequence[SSAValue | Operation],
+        outputs: Sequence[SSAValue | Operation],
         stride_patterns: ArrayAttr[StridePattern] | Sequence[StridePattern],
+        accelerator: StringAttr | str,
         body: Region,
     ) -> None:
         if not isinstance(stride_patterns, ArrayAttr):
             stride_patterns = ArrayAttr(stride_patterns)
+        if not isinstance(accelerator, StringAttr):
+            accelerator = StringAttr(accelerator)
         super().__init__(
             operands=[inputs, outputs],
             regions=[body],
-            properties={"stride_patterns": stride_patterns},
+            properties={"stride_patterns": stride_patterns, "accelerator": accelerator},
         )
 
     def verify_(self):
@@ -179,13 +183,13 @@ class StreamingRegionOp(IRDLOperation):
 
         streamer_config: StreamerConfiguration = streamer_interface.data
 
-        if len(self.stride_pattern) != streamer_config.size():
+        if len(self.stride_patterns) != streamer_config.size():
             raise VerifyException(
                 "Number of streamers does not equal number of stride patterns"
             )
 
         for stride_pattern, streamer in zip(
-            self.stride_pattern, streamer_config.streamers
+            self.stride_patterns, streamer_config.streamers
         ):
             if len(stride_pattern.temporal_strides) > streamer.temporal_dim:
                 raise VerifyException(
--- a/compiler/transforms/stream_snaxify.py
+++ b/compiler/transforms/stream_snaxify.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass
 
 from xdsl.dialects import builtin, memref, memref_stream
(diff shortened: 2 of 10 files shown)


compiler/dialects/snax_stream.py
```diff
@@ -142,17 +143,20 @@ class StreamingRegionOp(IRDLOperation):
...
+        inputs: Sequence[SSAValue | Operation],
+        outputs: Sequence[SSAValue | Operation],
         stride_patterns: ArrayAttr[StridePattern] | Sequence[StridePattern],
+        accelerator: StringAttr | str,
         body: Region,
     ) -> None:
         if not isinstance(stride_patterns, ArrayAttr):
             stride_patterns = ArrayAttr(stride_patterns)
+        if not isinstance(accelerator, StringAttr):
+            accelerator = StringAttr(accelerator)
```
JosseVanDelm, line 155: nit: I find positive `isinstance`s a bit more explicit.

tests/filecheck/transforms/stream-snaxify.mlir
```diff
@@ -0,0 +1,22 @@
+// RUN: ./compiler/snax-opt --split-input-file %s -p insert-accfg-op{accelerator=snax_alu},stream-snaxify | filecheck %s
```
JosseVanDelm, line 1: I'm wondering if it would make more sense to write the accfg-op explicitly here to make this test more standalone?
It does increase duplication a bit though :(

kernels/streamer_alu/streamer_add_stream.preprocfinal.mlir
```diff
@@ -19,18 +19,14 @@ func.func public @streamer_add(%arg0 : memref<?xi64, "L3">, %arg1 : memref<?xi64
...
-          "accelerator" = "snax_alu"}> ({
-    ^0(%103 : !stream.readable<i64>, %104 : !stream.readable<i64>, %105 : !stream.writable<i64>):
+    memref_stream.streaming_region {
+      patterns = [
+          #memref_stream.stride_pattern<ub = [4, 4], index_map = (d0, d1) -> (4 * d0 + d1)>,
+          #memref_stream.stride_pattern<ub = [4, 4], index_map = (d0, d1) -> (4 * d0 + d1)>,
+          #memref_stream.stride_pattern<ub = [4, 4], index_map = (d0, d1) -> (4 * d0 + d1)>
+      ]
+    } ins(%4, %7 : memref<?xi64, "L1">, memref<?xi64, "L1">) outs(%10 : memref<?xi64, "L1">) attrs = {accelerator="snax_alu"} {
+    ^bb0(%103: !stream.readable<i64>, %104: !stream.readable<i64>, %105: !stream.writable<i64>):
```
JosseVanDelm, line 29: unrelated (?) do we need a memory space attribute in the stream type?

compiler/transforms/stream_snaxify.py
```diff
@@ -0,0 +1,161 @@
...
+            module_op = module_op.parent_op()
+        if not module_op:
+            raise RuntimeError("Module Op not found")
+
+        trait = module_op.get_trait(SymbolTable)
+        assert trait is not None
+        acc_op = trait.lookup_symbol(module_op, accelerator_str)
+
+        if not isinstance(acc_op, AcceleratorOp):
+            raise RuntimeError("AcceleratorOp not found!")
```
JosseVanDelm, line 61: I believe I wrote a helper for this in accelerator.py? I think this one might be a bit more cautious, which I like. So maybe overwrite that one?

compiler/transforms/stream_snaxify.py
```diff
@@ -0,0 +1,161 @@
...
+
+        streamer_config = acc_op.attributes["streamer_config"]
+        if not isinstance(streamer_config, StreamerConfigurationAttr):
+            raise RuntimeError("Streamer interface not found for given accelerator op")
```
JosseVanDelm, line 65: I think one centralized helper for this would also be nice, since it was also in your other PR.

kernels/streamer_alu/streamer_add_stream.preprocfinal.mlir
```diff
@@ -55,26 +50,5 @@ func.func public @streamer_add(%arg0 : memref<?xi64, "L3">, %arg1 : memref<?xi64
...
-    "alu_mode" = 972 : i32, 
-    "loop_bound_alu" = 973 : i32}, 
-  "launch_fields" = {
-    "launch_streamer" = 970 : i32, 
-    "launch_alu" = 974 : i32}, 
-  "barrier" = 975 : i32}> {
-    "streamer_config" = #snax.streamer_config<r[1, 1], r[1, 1], r[1, 1]>
-} : () -> ()
-
 func.func private @snax_is_compute_core() -> i1
```
JosseVanDelm, line 53: Same comment as in other test, maybe it makes sense to keep this in? Not sure what you think is easiest

JosseVanDelm (approved):
This pass is very nice! I'm not sure we are testing this pass very thoroughly though.
If possible, can you maybe already explicitly add failling tests here for things that are not yet supported? We can move them to the supported tests later maybe? Since now the restrictions are very fresh in the mind, then they are very apparent if we hit a roadblock in one of the actual workloads later? What do you think?

kernels/streamer_alu/streamer_add_stream.preprocfinal.mlir
jorendumoulin: I don't think so, as a stream is associated with an accelerator. Maybe just a check when a memref is converted to a stream, that it is in the correct memory space for that accelerator.

compiler/transforms/stream_snaxify.py
jorendumoulin: A PR is coming up!

tests/filecheck/transforms/stream-snaxify.mlir
jorendumoulin: Hmm than the filecheck becomes really cluttered, and I don't see a downside to this method.

kernels/streamer_alu/streamer_add_stream.preprocfinal.mlir
JosseVanDelm: But if we are planning to use this dialect also for DMA regions we should add support for it, no?

tests/filecheck/transforms/stream-snaxify.mlir
JosseVanDelm: Yeah I agree!

compiler/dialects/snax_stream.py
JosseVanDelm: Thank you :heart:

PR #183, opened 2024-07-22
Use Transform Dialect with matmul

JonasCrols:
Add kernel test where simple_matmul script gets transformed using transform dialect to become half_tiled_matmul


runtime/Makefile.rules
```diff
@@ -74,9 +74,16 @@ LDFLAGS += -lsnRuntime
...
+
+MLIRTRANSFORM = --pass-pipeline='builtin.module(test-transform-dialect-interpreter{bind-first-extra-to-ops=linalg.quantized_matmul}, test-transform-dialect-erase-schedule)'
+
+%.mlir: %.transform.mlir
+	$(MLIROPT) $(MLIRTRANSFORM) -o $@ $<
+
 # MLIR Preprocessing
 
-MLIRPREPROCFLAGS = --pass-pipeline='builtin.module(func.func(tosa-to-linalg-named, tosa-to-tensor, tosa-to-scf, tosa-to-linalg))'
+MLIRPREPROCFLAGS = --pass-pipeline='builtin.module( func.func(tosa-to-linalg-named, tosa-to-tensor, tosa-to-scf, tosa-to-linalg))'
```
JosseVanDelm, line 86: Not sure why this change is necessary.

kernels/transform_matmul/matmul.transform.mlir
```diff
@@ -0,0 +1,19 @@
...
+
+transform.sequence failures(propagate) {
+^bb0(%arg0: !transform.any_op,
+     %arg1: !transform.op<"linalg.quantized_matmul">):
+  // The actual tiling transformation takes tile sizes as attributes.
+   %loop1, %loop2, %tiled = transform.structured.tile %arg1 [8, 8]
+    : (!transform.op<"linalg.quantized_matmul">) -> (!transform.any_op, !transform.any_op, !transform.any_op)
+  transform.yield
+}
+
```
JosseVanDelm, line 16: If this is the only line that is changing alongside the added options to run the compiler, we might just want to add this to another folder instead, of adding yet another entire 200 lines. @jorendumoulin what do you think?

JosseVanDelm (approved):
Very nice :heart: :heart_eyes_cat: :heart_hands:

kernels/transform_matmul/Makefile
```diff
@@ -0,0 +1,40 @@
+# Courtesy of Federico Ficarelli
+
+.DEFAULT_GOAL := all
+
+include ../../runtime/snax-gemm.rules
```
jorendumoulin, line 5: Let's use the streamer gemm?

jorendumoulin (approved):
Cool!

I agree with Josse - Let's just put this in the `streamer_gemm` folder. Only the mlir file is different I think?

PR #189, opened 2024-07-26
Add reuse-memref-space pass

JonasCrols:
Add new reuse-memref-space pass that moves allocations outside for-loops when possible. Also added filecheck


compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,254 @@
...
+            Check if the Dim operation can be moved outside the loop.
+            """
+            if before_loop(dim):
+                return True
+            subview = dim.source.owner
+            if isinstance(subview, Block):
+                # This happens when the dim is called on an input argument
+                return True
+            assert isinstance(subview, memref.Subview)
+            index = int(dim.index.owner.value.value.data)
```
jorendumoulin, line 62: There is a third case we should handle here:
If there are two matmuls after each other, like
```
A = B * C
F = A * G
```
For the calculation of `F = A * G`, we need to allocate space for A. A is not the result of a subview, but instead, the result of a previous `alloc` operation.

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,254 @@
...
+                        ops_to_add.append(new_dim)
+                        dynamic_sizes.append(new_dim.results[0])
+                else:
+                    new_constant = get_constant_value(size.owner)
+                    ops_to_add.append(new_constant)
+                    dynamic_sizes.append(new_constant.results[0])
+            return dynamic_sizes
+
+        # Repeat When Changes are made.
+        changes_made = False
```
jorendumoulin, line 216: You don't need to return anything in this function, xdsl checks if changes are made based on if the rewriter does something

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,254 @@
...
+from xdsl.rewriter import InsertPoint
+
+
+class MoveMemrefAllocations(RewritePattern):
+    """
+    This class represents a rewrite pattern for moving memref allocations outside
+    a double for-loop. This is possible when each loop a new memref-space is allocated with
+    identical sizes and the space is not used outside the loop.
+    """
+
```
jorendumoulin, line 24: This is very nice, but this rewritepattern is too complex. I think it makes a lot of sense to split this up in multiple patterns. These can then be combined in one large pass using for example like this:
```python
        PatternRewriteWalker(
            GreedyRewritePatternApplier(
                [
                    Pattern1(),
                    Pattern2(),
                    Pattern3(),
                    ...
                ]
            ),
            apply_recursively=False,
        ).rewrite_module(op)
```
This will then apply all the patterns recursively until nothing changes anymore.
If I'm correct, this pass now applies 3 different patterns:

- If an operation is inside a for loop, and has no side effects (hase the `Pure` trait), and does not depend on anything inside the for loop, it can be hoisted out of the for loop.
- Special case for the dim op -> if it depends on subviews / min, and is only used by allocs, make the approximation, and hoist it out of the for loop.
- Special case for the alloc op -> even though this operation has side effects, we still want to hoist it out of the for loop if it does not depend on anything in the for loop, such that the resulting buffer is reused.

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,254 @@
...
+            """
+            Find the parent for-loop of the operation.
+            If no for-loop is found, return None.
+            """
+            op = op.parent_op()
+            while not isinstance(op, scf.For):
+                if op.parent_op() is None:
+                    return None
+                op = op.parent_op()
+            return op
```
jorendumoulin, line 37: These are all nice helper functions, but define them outside of the class if they do not depend on it!

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,254 @@
...
+            """
+            return find_parent_for_loop(op) is not find_parent_for_loop(alloc_op)
+
+        def is_in_loop(op: Operation) -> bool:
+            """
+            Check if the operation is inside a loop.
+            """
+            return find_parent_for_loop(op) is not None
+
+        def can_move_dim(dim: memref.Dim) -> bool:
```
jorendumoulin, line 51: this feels like a separate rewrite pattern

tests/filecheck/transforms/reuse-memref-space.mlir
```diff
@@ -0,0 +1,319 @@
...
+//CHECK-NEXT:     %1 = arith.constant 1 : index
+//CHECK-NEXT:     %2 = arith.constant 0 : i32
+//CHECK-NEXT:     %3 = arith.constant 0 : index
+//CHECK-NEXT:     %4 = arith.constant 16 : index
+//CHECK-NEXT:     %5 = arith.constant 8 : index
+//CHECK-NEXT:     %6 = memref.alloc(%4, %5) {"alignment" = 64 : i64} : memref<?x?xi8, #tsl.tsl<[?, 8] -> (?, 8), [?, 8] -> (256, 1)>, "L1">
+//CHECK-NEXT:     scf.for %arg2 = %3 to %0 step %1 {
+//CHECK-NEXT:       %7 = arith.constant 16 : index
+//CHECK-NEXT:       %8 = arith.constant 8 : index
+//CHECK-NEXT:     }
```
jorendumoulin, line 59: Instead of creating new operations, I think it makes more sense to hoist the constants outside of the loop too?

jorendumoulin (approved):
Thank you! This is already very nice :)

The main things I'd like to change are this:

- Instead of redefining constant operations outside of the loop, just move them outside of the loop.
- Splitting the single rewrite pattern into more smaller, logical patterns. This will make it much clearer to reason about things when we hit a possible error in the future :)

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,254 @@
...
+    op_type_rewrite_pattern,
+)
+from xdsl.rewriter import InsertPoint
+
+
+class MoveMemrefAllocations(RewritePattern):
+    """
+    This class represents a rewrite pattern for moving memref allocations outside
+    a double for-loop. This is possible when each loop a new memref-space is allocated with
+    identical sizes and the space is not used outside the loop.
```
JosseVanDelm, line 22: A few things come to mind reading this:
* Why does this mention a double-for loop? Does it not work for double loops? Maybe we should state that it works for `nested loops`?
* When talking about memref-space, do you mean the space associated with the storage of a memref? If that is the case I would just refer to it as `the memref`. Now it is a bit misleading, since memrefs also have a memory space associated with them, which does not mean the space, but rather a level in memory, like "L3" or "LLC".

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,254 @@
...
+        def replace_dim(dim: memref.Dim) -> Operation:
+            """
+            Replace Dim operation with either Dim outside the loop, or a new Constant.
+            """
+            if before_loop(dim):
+                return dim
+            subview = dim.source.owner
+            if isinstance(subview, Block):
+                # This happens when the dim is called on an input argument
+                return dim
```
JosseVanDelm, line 167: nit: it is slightly confusing that at this point you don't know it is a subview yet :smile:

tests/filecheck/transforms/reuse-memref-space.mlir
```diff
@@ -0,0 +1,319 @@
...
+        %4 = "affine.min"(%1, %0) <{"map" = affine_map<(d0)[s0] -> (8, ((d0 * -1) + s0))>}> : (index, index) -> index
+        %5 = "memref.dim"(%arg0, %1) :(memref<?x?xi8, "L3">, index) -> index
+        %6 = memref.alloc(%4, %5) {"alignment" = 64 : i64} : memref<?x?xi8, #tsl.tsl<[?, 8] -> (?, 8), [?, 8] -> (256, 1)>, "L1">
+        }
+    func.return
+  }
+}
+
+// Check if the memref.alloc is moved out of the loop when affine.min is inside loop
+// so that a new arith.constant is created equal to maximum of affine.min
```
JosseVanDelm, line 178: ~I feel like doing something like "unrolling" your affine min's should not be part of this pass. Can't you just hoist the affine.min outside of the loop as well?~
Okay it is actually defined as a loop carried variable (so you can not just hoist it out), but that is not the case in this test?

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,254 @@
...
+                symbol_operands=alloc_op.symbol_operands,
+                result_type=alloc_op.results[0].type,
+                alignment=alloc_op.alignment,
+            )
+            rewriter._replace_all_uses_with(
+                # This is a private function of Rewriter Class and should be replaced with a public function,
+                # but for now it works
+                alloc_op.results[0],
+                new_alloc_op.results[0],
+            )
```
JosseVanDelm, line 234: Can you not simply use `replace_op` instead?

JosseVanDelm (commented):
Hi Jonas, this is an amazing PR! But I think with comments like Joren's we can make it even more amazing.

I agree with @jorendumoulin's suggestion to split it out in multiple parts. That way it is easier to see what is going on, and it is likely that some of your work is useful for later passes, or even for upstreaming to xdsl.

it seems that some of this stuff already is part of an `mlir-opt` pass, but we probably can not use it because we are using custom attributes on the memref type. So I'm just putting this here for later reference
```mlir
builtin.module {
  func.func @streamer_matmul(%arg0 : memref<?x?xi8, "L3">, %arg1 : memref<?x?xi32, "L3">) {
    %0 = arith.constant 12 : index
    %1 = arith.constant 1 : index
    %2 = arith.constant 0 : i32
    %3 = arith.constant 0 : index
    scf.for %arg2 = %3 to %0 step %1{
        %4 = arith.constant 16 : index
        %5 = arith.constant 8 : index
        %6 = memref.alloc(%4, %5) {"alignment" = 64 : i64} : memref<?x?xi8, "L1">
        "test.op"(%6) : (memref<?x?xi8, "L1">) -> ()
        }
    func.return
  }
}
```
with `--cse` and `--loop-invariant-code-motion`:
```mlir
module {
  func.func @streamer_matmul(%arg0: memref<?x?xi8, "L3">, %arg1: memref<?x?xi32, "L3">) {
    %c12 = arith.constant 12 : index
    %c1 = arith.constant 1 : index
    %c0 = arith.constant 0 : index
    %c16 = arith.constant 16 : index
    %c8 = arith.constant 8 : index
    scf.for %arg2 = %c0 to %c12 step %c1 {
      %alloc = memref.alloc(%c16, %c8) {alignment = 64 : i64} : memref<?x?xi8, "L1">
      "test.op"(%alloc) : (memref<?x?xi8, "L1">) -> ()
    }
    return
  }
}
```
Thanks!

compiler/transforms/reuse_memref_space.py
jorendumoulin: Yes I think all functions starting with `_` are not meant to be used by the outside world

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,254 @@
...
+            Operations can be moved if they comply with the following conditions:
+            1. The operation is inside a loop.
+            2. All operands are defined outside the loop.
+            3. The operation is Pure, it doesnt have any side effects.
+            """
+            if all(
+                [
+                    is_in_loop(op),
+                    defined_outside_loop(op),
+                    Pure() in op.traits or isinstance(op, memref.Alloc),
```
JonasCrols, line 70: I decided not to split the rewrite patterns for the alloc case and the general case as all almost code is equal. The only difference lies here in the testing if it can be moved, so I just worked with an or-operation to cover both cases. If we were to push the general case upstream to xDSL, the alloc case needs te be removed

jorendumoulin: Ok! Makes sense

compiler/transforms/reuse_memref_space.py
```diff
@@ -1,72 +1,104 @@
...
+    Check if all operands of the operation are defined outside the loop.
+    """
+    for operand in op.operands:
+        if find_parent_for_loop(operand.owner) is find_parent_for_loop(op):
+            return False
+    return True
 
 
-class MoveMemrefAllocations(RewritePattern):
+class LowerPureOperations(RewritePattern):
```
JosseVanDelm, line 50: I think you are referring to loop hoisting, or loop-invariant-code-motion. Lowering is something different.
```suggestion
class LoopHoistPureOperations(RewritePattern):
```

compiler/transforms/reuse_memref_space.py
JosseVanDelm: @JonasCrols, maybe instead, make the memref.Alloc a parameter of this rewrite pattern? Then you can make it more general? I believe I've seen @AntonLydike do this a couple of times.

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,284 @@
...
+    for operand in op.operands:
+        if find_parent_for_loop(operand.owner) is find_parent_for_loop(op):
+            return False
+    return True
+
+
+class LowerPureOperations(RewritePattern):
+    """
+    This class represents a rewrite pattern for moving any Operation outside
+    a (nested) for-loop. An Oparation can be moved if it is pure, that is, it
```
JosseVanDelm, line 53:
```suggestion
    a (nested) for-loop. An Operation can be moved if it is pure, that is, it
```

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,284 @@
...
+            for use in dim_op.results[0].uses:
+                if not isinstance(use.operation, memref.Alloc) and not isinstance(
+                    use.operation, memref.Subview
+                ):
+                    return True
+            return False
+
+        def can_move_dim(dim_op: memref.Dim) -> bool:
+            """
+            Allocs can be moved if they comply with the following conditions:
```
JosseVanDelm, line 162:
```suggestion
        def can_move_dim(dim_op: memref.Dim) -> bool:
            """
            Dims can be moved if they comply with the following conditions:
```

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,284 @@
...
+                    return dimension_outside_loop(new_op)
+                else:
+                    return False
+            if isinstance(memref_op, linalg.MatmulOp):
+                new_memref_op = memref_op.inputs[index].owner
+                return memref_op_outside_loop(new_memref_op, index)
+            # TODO: Add support for other memref operations, like matmul
+            return False
+
+        def used_by_not_alloc_subview(dim_op: memref.Dim) -> bool:
```
JosseVanDelm, line 149: This reads a bit easier IMHO
```suggestion
        def used_by_neither_alloc_nor_subview(dim_op: memref.Dim) -> bool:
```

compiler/transforms/reuse_memref_space.py
```diff
@@ -0,0 +1,284 @@
...
+                if isinstance(new_op, arith.Constant) or isinstance(
+                    new_op, affine.MinOp
+                ):
+                    return new_op
+                else:
+                    return get_new_dim_op(new_op)
+            if isinstance(memref_op, linalg.MatmulOp):
+                new_memref_ssa = memref_op.inputs[index]
+                return memref_op_outside_loop(new_memref_ssa, index_constant)
+            # TODO: Add support for other memref operations, like matmul
```
JosseVanDelm, line 237: Can we not do this on generic linalg operations instead?

JosseVanDelm (commented):
I have a few more possibly dumb comments

compiler/transforms/reuse_memref_space.py
JonasCrols: That would not be necessary as the Linalg.generic operation uses an already defined memref (from either an alloc ror a subview) which would already be lowered if possible

PR #192, opened 2024-08-06
util: add canonicalize_affine_map

jorendumoulin:
The algorithms used in the stream transform passes try to compare different affine maps. To be able to compare them, this utility function tries to implement some canonicalization. It is quite barebones, but it suffices for now. Future work is to implement the full flattening of affine maps in upstream MLIR in xdsl so we can use that, but this is quite some work.

### DIFF
--- a/compiler/util/canonicalize_affine.py
+++ b/compiler/util/canonicalize_affine.py
@@ -1,0 +1,111 @@
+from xdsl.ir.affine import (
+    AffineBinaryOpExpr,
+    AffineBinaryOpKind,
+    AffineConstantExpr,
+    AffineDimExpr,
+    AffineExpr,
+    AffineMap,
+)
+
+
+def get_dim(expr: AffineExpr) -> int | None:
+    if isinstance(expr, AffineDimExpr):
+        return expr.position
+    if isinstance(expr, AffineBinaryOpExpr):
+        result = get_dim(expr.lhs)
+        if result is not None:
+            return result
+        result = get_dim(expr.rhs)
+        if result is not None:
+            return result
+
+
+def canonicalize_addition(expr: AffineBinaryOpExpr) -> AffineExpr:
+    assert expr.kind is AffineBinaryOpKind.Add
+    if isinstance(expr.lhs, AffineConstantExpr):
+        # lhs is constant
+        if isinstance(expr.rhs, AffineConstantExpr):
+            # both are constants
+            return AffineConstantExpr(expr.lhs.value + expr.rhs.value)
+        else:
+            # move constant to rhs
+            expr = AffineBinaryOpExpr(expr.kind, expr.rhs, expr.lhs)
+    if isinstance(expr.rhs, AffineConstantExpr):
+        # rhs is constant, lhs is not
+        if expr.rhs.value == 0:
+            return expr.lhs
+    # order by minimum dimension first
+    dim_lhs = get_dim(expr.lhs)  # 3
+    dim_rhs = get_dim(expr.rhs)  # 0
+    if dim_rhs is not None:
+        if dim_lhs is None or dim_lhs > dim_rhs:
+            return expr.rhs + expr.lhs
+    return expr
+
+
+def canonicalize_multiplication(expr: AffineBinaryOpExpr) -> AffineExpr:
+    assert expr.kind is AffineBinaryOpKind.Mul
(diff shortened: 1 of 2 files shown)


compiler/util/canonicalize_affine.py
```diff
@@ -0,0 +1,121 @@
...
+    AffineMap,
+)
+
+
+def get_dim(expr: AffineExpr) -> int | None:
+    # get the dimension of this expression. If d0 appears, return 0.
+    # If d1 appears, return 1. This will fail for nested additions in
+    # which multiple dimension expressions appear. This happens quite
+    # often, but can only be solved with a full flattening of the affine map.
+    # this is todo
```
JosseVanDelm, line 16:
```suggestion
    """
    get the dimension of this expression. If d0 appears, return 0.
    If d1 appears, return 1. This will fail for nested additions in
    which multiple dimension expressions appear. This happens quite
    often, but can only be solved with a full flattening of the affine map.
    this is todo
    """
```

compiler/util/canonicalize_affine.py
```diff
@@ -0,0 +1,121 @@
...
+        if expr.rhs.value == 0:
+            return expr.lhs
+    # order by minimum dimension first
+    # d0 + d1: good. d1 + d0: bad
+    dim_lhs = get_dim(expr.lhs)
+    dim_rhs = get_dim(expr.rhs)
+    if dim_rhs is not None:
+        if dim_lhs is None or dim_lhs > dim_rhs:
+            return expr.rhs + expr.lhs
+    return expr
```
JosseVanDelm, line 51: Can you move most of these comments together into a docstring that says everything this function does, instead of it being scattered across this function? I'm giving an example below

compiler/util/canonicalize_affine.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def canonicalize_multiplication(expr: AffineBinaryOpExpr) -> AffineExpr:
+    # always put the constant on rhs
```
JosseVanDelm, line 55:
```suggestion
    """
    Canonicalize multiplication. If both are constants return the constant folded multiplication.
    If one of them is constant move it to the right hand side.
    Omit multiplications with a constant of 1.
    """
```

compiler/util/canonicalize_affine.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+    if expr.kind is AffineBinaryOpKind.Add:
+        return canonicalize_addition(expr)
+    if expr.kind is AffineBinaryOpKind.Mul:
+        return canonicalize_multiplication(expr)
+    if expr.kind is AffineBinaryOpKind.FloorDiv:
+        return canonicalize_floordiv(expr)
+    if expr.kind is AffineBinaryOpKind.Mod:
+        return canonicalize_mod(expr)
+    return expr
```
JosseVanDelm, line 104:
```suggestion
    match expr.kind:
        case AffineBinaryOpKind.Add:
            return canonicalize_addition(expr)
        case AffineBinaryOpKind.Mul:
            return canonicalize_multiplication(expr)
        case AffineBinaryOpKind.FloorDiv:
            return canonicalize_floordiv(expr)
        case AffineBinaryOpKind.Mod:
            return canonicalize_mod(expr)
        case _:
            return expr
```

compiler/util/canonicalize_affine.py
```diff
@@ -0,0 +1,121 @@
...
+def canonicalize_expr(expr: AffineExpr) -> AffineExpr:
+    if isinstance(expr, AffineBinaryOpExpr):
+        return canonicalize_binary_op(expr)
+
+    return expr
+
+
+# helper function to canonicalize affine maps
+def canonicalize_map(map: AffineMap) -> AffineMap:
+    # canonicalize each result and construct new affine map
```
JosseVanDelm, line 116: I'm not sure what this means :(

JosseVanDelm (approved):
Overall this is super nice PR! If possible would like to have a bit more of a consistent commenting style please :heart:

compiler/util/canonicalize_affine.py
jorendumoulin: An affine map is built out of multiple affine expressions. These are called `results`. So we canonicalize an affine map by canonicalizing all o the expressions / results.

PR #193, opened 2024-08-06
util: (kernel_type) add Kerneltype.ADD kernel

jorendumoulin:
The snax-alu accelerator implements elementwise addition. To be able to dispatch and detect this, the ADD kernel type is added to the kernel utilities.

This also fixes some weird things I found in the test.
Such as `arith.Constant(IndexType(), 1)`, which is a completely invalid call of the constructor, but does not throw an error as the op is never verified/inserted in ir/ value is requested :)

### DIFF
--- a/compiler/util/kernel_type.py
+++ b/compiler/util/kernel_type.py
@@ -38,12 +38,13 @@ class KernelType(Enum):
             out += (a - zp_a) * (b - zp_b)
     """
 
+    ADD = "add"
     MUL = "mul"
     MAC = "mac"
     QMAC = "qmac"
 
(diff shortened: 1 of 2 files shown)


compiler/util/kernel_type.py
```diff
@@ -38,12 +38,13 @@ class KernelType(Enum):
             out += (a - zp_a) * (b - zp_b)
     """
 
+    ADD = "add"
```
JosseVanDelm, line 41: Also add comment to your nice docstring :)

JosseVanDelm (approved):
:heart:

PR #194, opened 2024-08-06
transform: (dispatch_kernels) add SnaxALU Rules

jorendumoulin:
This, so the dispatching now dispatches elementwise additions to the snax alu. 

Additionally, this pass decides if the kernel can be dispatched using the stream programming model. 
If this is the case the `_stream` suffix is appended to the library call. This is a temporary workaround such that the `convert-linalg-to-memref-stream` only converts the generic kernels dispatched to a streamer accelerator.

As this only supports static sizes for now, dynamic sizes still use the old programming model.

### DIFF
--- a/compiler/transforms/dispatch_kernels.py
+++ b/compiler/transforms/dispatch_kernels.py
@@ -1,6 +1,7 @@
 from xdsl.context import MLContext
 from xdsl.dialects import builtin, linalg
 from xdsl.dialects.memref import MemRefType
+from xdsl.ir import Operation
 from xdsl.ir.affine import AffineDimExpr, AffineMap
 from xdsl.passes import ModulePass
 from xdsl.pattern_rewriter import (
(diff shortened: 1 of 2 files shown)


tests/filecheck/transforms/dispatch_kernels.mlir
```diff
@@ -57,3 +57,55 @@
...
+// CHECK-NEXT:    "func.func"() <{"function_type" = (memref<16xi64>, memref<16xi64>, memref<16xi64>) -> (), "sym_name" = "streamer_add", "sym_visibility" = "public"}> ({
+// CHECK-NEXT:    ^0(%arg0 : memref<16xi64>, %arg1 : memref<16xi64>, %arg2 : memref<16xi64>):
+// CHECK-NEXT:      "linalg.generic"(%arg0, %arg1, %arg2) <{"indexing_maps" = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>], "iterator_types" = [#linalg.iterator_type<parallel>], "operandSegmentSizes" = array<i32: 2, 1>, "library_call" = "snax_alu_stream"}> ({
+// CHECK-NEXT:      ^1(%arg3 : i64, %arg4 : i64, %arg5 : i64):
+// CHECK-NEXT:        %0 = "arith.addi"(%arg3, %arg4) : (i64, i64) -> i64
+// CHECK-NEXT:        "linalg.yield"(%0) : (i64) -> ()
+// CHECK-NEXT:      }) : (memref<16xi64>, memref<16xi64>, memref<16xi64>) -> ()
+// CHECK-NEXT:      "func.return"() : () -> ()
+// CHECK-NEXT:    }) : () -> ()
+// CHECK-NEXT:  }) : () -> ()
```
JosseVanDelm, line 85: Maybe only check the library_call attribute? Also, maybe you can just check if it is _not_ set to snax_alu_stream instead?

compiler/transforms/dispatch_kernels.py
```diff
@@ -13,6 +13,46 @@
 from compiler.util.kernel_type import KernelType
 
 
+class DispatchSnaxALU(RewritePattern):
```
JosseVanDelm, line 16:
```suggestion
class DispatchSnaxALU(RewritePattern):
     """
     Dispatch eligible Linalg Generic ops to SNAX ALU accelerator by altering the library_call attribute
     """
```

compiler/transforms/dispatch_kernels.py
```diff
@@ -129,3 +169,4 @@ class DispatchKernels(ModulePass):
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
         PatternRewriteWalker(DispatchElementwiseMult()).rewrite_module(op)
         PatternRewriteWalker(DispatchQMatMul()).rewrite_module(op)
+        PatternRewriteWalker(DispatchSnaxALU()).rewrite_module(op)
```
JosseVanDelm, line 172: The naming you introduce here is better, but now it is a bit inconsistent with the previous other passes.

I'm wondering if we should co-locate these dispatch passes with their accelerator implementations in the accelerators folder?

JosseVanDelm (approved):
Looks good!
I have a few (future) suggestions that don't necessary need to be implemented, but I'd be interested to learn what you think :)

tests/filecheck/transforms/dispatch_kernels.mlir
jorendumoulin: yes, because it checks for "snax_alu", the quotations are checked as well, so snax_alu_stream is impossible

compiler/transforms/dispatch_kernels.py
jorendumoulin: tbh this entire pass needs a rethink and do-over. I hope we can engineer a better system once the accelerator templates start to take better form

PR #196, opened 2024-08-06
transform: (guarded-linalg-to-memref-stream) init pass

jorendumoulin:
Upstream xdsl converts all linalg.generic calls to memref.generic calls. We don't always want this. This pass adds an extra guard to only convert linalg.generic which are dispatched to an accelerator, appended with the _stream suffix. It then removes the suffix after the conversion is done.

### DIFF
--- a/compiler/transforms/guarded_linalg_to_memref_stream.py
+++ b/compiler/transforms/guarded_linalg_to_memref_stream.py
@@ -9,11 +9,13 @@ from xdsl.pattern_rewriter import (
     RewritePattern,
     op_type_rewrite_pattern,
 )
-from xdsl.transforms.convert_linalg_to_memref_stream import ConvertGenericOpPattern, ConvertYieldOpPattern
+from xdsl.transforms.convert_linalg_to_memref_stream import (
+    ConvertGenericOpPattern,
+    ConvertYieldOpPattern,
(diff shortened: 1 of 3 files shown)


compiler/transforms/guarded_linalg_to_memref_stream.py
```diff
@@ -0,0 +1,44 @@
...
+)
+
+
+class GuardedGenericOpPattern(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: linalg.Generic, rewriter: PatternRewriter) -> None:
+        """
+        This pattern is a guarded version of the pattern in upstream xDSL.
+        This is required because we can't (or don't want to) lower all
+        generic calls through the stream dialects.
```
JosseVanDelm, line 24:
```suggestion
        generic calls through the stream dialects.
        It only matches on linalg.generic's that have library call defined with the name of an accelerator that ends with _stream.
        This pass removes the `_stream` suffix from converted linalg.generics
```

JosseVanDelm (approved):
I guess an upstream pass with a guard option would be a tad bit nicer, but this is a very good alternative for our current purposes!
Thanks @jorendumoulin!

PR #197, opened 2024-08-06
dialects: (accfg) fix streamer lowering

jorendumoulin:
The order of programming is first streamer, then dim. This was implemented the other way around (wrong)


JosseVanDelm (approved):
Allright :cowboy_hat_face: !

PR #198, opened 2024-08-06
dialects: (tsl) implement get_affine_map

jorendumoulin:
To be able to stream-snaxify, we must have an affine mapping from data to memory. This implements this for TSL layouts, such that these are supported as well as NoneAttr and StridedLayoutAttr.

One caveat is that the definition of `get_affine` is to return the element offset. TSL returns the byte offset. In hindsight, I will probably refactor tsl layouts to use element offset instead at a later stage, once the dust settles.

For a TSL with bounds

$$
    \left[b_{0,M}, ..., b_{0,1}, b_{0, 0}\right], \left[b_{1,K}, ..., b_{1,1}, b_{1, 0}\right], ...
    $$

and tiled strides for every dim $d$:

$$
    \left[b_{d,M}, ..., b_{d,1}, b_{d, 0}\right] \rightarrow \left(s_{d,M}, ..., s_{d,1}, s_{d, 0}\right)
$$

The affine map is given by:

$$(i_0, i_1, ..., i_n) \rightarrow
\sum_{n=0}^{N-1}\sum_{m=0}^{M-1} s_{n,m} \cdot \left\lfloor\left( i_n \bmod \prod_{k=0}^{m}b_{n,k} \right) \div \prod_{k=0}^{m-1}b_{n,k}\right\rfloor$$

### DIFF
--- a/compiler/dialects/tsl.py
+++ b/compiler/dialects/tsl.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
 from math import prod
 
 from xdsl.dialects.arith import Constant, DivUI, Muli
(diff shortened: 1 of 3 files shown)


compiler/dialects/tsl.py
```diff
@@ -38,7 +38,28 @@ def print_parameter(self, printer: Printer) -> None:
         printer.print_string(f"<{self.data}>")
 
     def get_affine_map(self) -> AffineMap:
-        raise NotImplementedError("Still to do!")
+        if self.data.is_dynamic():
+            raise NotImplementedError("Dynamic case is not implemented yet!")
+
+        # TODO: the affine map should result in element offset, not byte offset
+        # i will probably transition the tsl definition to element offset
+        # as well, to make everything more convenien
```
JosseVanDelm, line 46:
```suggestion
        # as well, to make everything more convenient
```

JosseVanDelm (approved):
Can you please add these very pretty formulas to your tsl readme?
Would also be nice if you could exlicitly elaborate a bit more on the meaning of all symbols (N, M, s, d, ...) :)

PR #201, opened 2024-08-13
Add gemmini Loop WS kernel support through ACCFG

JosseVanDelm:
This PR implements the following:
* A conversion from linalg generic to Gemmini's weight-stationary kernel
* A "temporary fix" to allow deduplicated RoCC code to be converted to CSRs.
Right now there is no way to create pairs if one of the pairs is deduplicated. 
The `create_pairs` function will try to catch the previous state to create a pair again, but this is not possible if only half of the pair is set in said previous state.
As a first fix, this test assumes all settings are set in pairs, and a constant default is appended to the very first setup op in the state chain (the one without an in_state). This makes sure that a default value is set to rs1 if rs2 is set - but not rs1 - and vice versa.  A more thorough fix for this actually might reset all values at first? Any suggestions welcome.
- [x] add test for this ^^

### DIFF
--- a/kernels/gemmini/Makefile
+++ b/kernels/gemmini/Makefile
@@ -13,9 +13,7 @@ CFLAGS += -Wall -Wextra
 SNAXOPTACCFLAGS = -p insert-accfg-op{accelerator=gemmini},convert-linalg-to-accfg,convert-accfg-to-csr
 
 tiled_matmul.mlir: tiled_matmul.transform.mlir
-	# Use quick patch to remove transform module
-	${MLIROPT} --pass-pipeline='builtin.module(transform-interpreter, test-transform-dialect-erase-schedule)' $^ | sed '/^  module @transforms attributes {transform.with_named_sequence} {$$/,/^  }$$/d' > $@
-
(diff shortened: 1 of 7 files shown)


JosseVanDelm:
Hi! I'm opening this for review.
I understand the current version of the container does not support the current transform Dialect statements as per @JonasCrols' comments.

JosseVanDelm:
> (I can also hack a sed fix in there if required. 

This is what I went with, since it works pretty well, also over at https://github.com/KULeuven-MICAS/gemmini-rocc-tests/blob/dev/bareMetalMLIR

kernels/gemmini/Makefile
```diff
@@ -0,0 +1,42 @@
...
+%.ll.mlir: %.postproc.mlir
+	${MLIROPT} --test-lower-to-llvm -o $@ $<
+
+%.ll: %.postproc.mlir
+	${MLIRTRANSLATE} --mlir-to-llvmir -o $@ $<
+
+%.x: %.ll
+	${CC} -x ir $< -c -O3 -o $@ -Wno-override-module --target=riscv64-unknown-elf -mcpu=generic-rv64 -march=rv64gc
+
+allrun: all
```
JosseVanDelm, line 33: Note that it does not actually run this, otherwise we'd have to install spike with gemmini extensions

jorendumoulin (approved):
Okay!

PR #204, opened 2024-08-26
transform: (schedule-memref-linalg pass)

jorendumoulin:
This PR adds the first version of the schedule-memref-linalg pass.
This pass matches and transorms an operation on an accelerator template.

At this stage, this is quite barebones, and only functional for easy exemples (elementwise operation on elementwise snax-alu accelerator/ gemm operation on gemm accelerator).

This document is an effort to explain its workings at the current stage.
[schedule-memref-linalg.pdf](https://github.com/user-attachments/files/16744943/schedule-memref-linalg.pdf)

Many improvements to this algorithm will come very shortly :)


jorendumoulin:
i have the wrong code here

PR #205, opened 2024-08-26
transform: schedule-memref-linalg pass

jorendumoulin:
This PR adds the first version of the schedule-memref-linalg pass.
This pass matches and transforms an operation on an accelerator template.

At this stage, this is quite barebones, and only functional for easy examples (elementwise operation on elementwise snax-alu accelerator/ gemm operation on gemm accelerator).

This document is an effort to explain its workings at the current stage.
[schedule-memref-linalg.pdf](https://github.com/user-attachments/files/16744943/schedule-memref-linalg.pdf)

Many improvements to this algorithm will come very shortly :)

### DIFF
--- a/compiler/transforms/schedule_memref_linalg.py
+++ b/compiler/transforms/schedule_memref_linalg.py
@@ -114,6 +114,7 @@ class ScheduleMemrefLinalgRewriter(RewritePattern):
     Takes a memref_stream.generic op as input and wraps it in a memref_stream streaming region op
     to determine the access patterns, the operation is scheduled to an accelerator template.
     """
+
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: memref_stream.GenericOp, rewriter: PatternRewriter):
         if any(isinstance(operand.type, stream.StreamType) for operand in op.operands):
--- a/compiler/transforms/schedule_memref_linalg.py
+++ b/compiler/transforms/schedule_memref_linalg.py
@@ -16,7 +16,7 @@ from compiler.util.canonicalize_affine import canonicalize_map
 
 def disable_dims(map: AffineMap, dim: int) -> AffineMap:
     """
-    Returns an affine map with the first `dim` dimensions set to 0
+    Returns an affine map with the leftmost `dim` dimensions set to 0
 
     For example:
         (d0, d1, d2) -> d0 + d1 + d2
@@ -38,7 +38,7 @@ def disable_dims(map: AffineMap, dim: int) -> AffineMap:
 
 def rotate_dims(map: AffineMap, dim: int) -> AffineMap:
     """
-    Returns an affine map with the first `dim` dimensions rotated
+    Returns an affine map with the leftmost `dim` dimensions rotated
 
     For example:
         (d0, d1, d2) -> 1 * d0 + 2 * d1 + 3 * d2
@@ -66,10 +66,10 @@ def rotate_bounds(bounds: list[int | None], dim: int) -> list[int | None]:
     return bounds
 
 
-def split_dim(map: AffineMap, dim: int, template_bound: int) -> AffineMap:
+def tile_dim(map: AffineMap, dim: int, template_bound: int) -> AffineMap:
     """
     Returns the bounds and a new affine map with the `dim` dimension split up into two
-    This translates to creating two for loops with adjuste bounds from one for loop
+    This translates to creating two for loops with adjusted bounds from one for loop
 
 
     For example:
@@ -91,11 +91,11 @@ def split_dim(map: AffineMap, dim: int, template_bound: int) -> AffineMap:
     return result
 
 
-def split_bounds(
+def tile_bounds(
     bounds: list[int | None], dim: int, template_bound: int
 ) -> list[int | None]:
     """
-    Returns the bounds after applying `split_dim` in similar fashion.
+    Returns the bounds after applying `tile_dim` in similar fashion.
 
     For example:
         [2, 8, 2]
@@ -125,7 +125,9 @@ class ScheduleMemrefLinalgRewriter(RewritePattern):
             isinstance(op.library_call, builtin.StringAttr)
             and op.library_call.data in ("snax_alu", "snax_gemm")
         ):
-            raise NotImplementedError("panic!")
+            raise NotImplementedError(
+                "only snax_alu and snax_gemm are supported right now"
+            )
 
         # only handle snax_alu and snax_gemm for now
         if op.library_call.data == "snax_alu":
@@ -139,8 +141,6 @@ class ScheduleMemrefLinalgRewriter(RewritePattern):
                 AffineMap(6, 0, (M * 8 + m, N * 8 + n)),
             ]
             template_bounds = (None, None, None, 8, 8, 8)
-        else:
-            raise RuntimeError("panic!")
 
         # only take patterns of memref operands
         schedule = [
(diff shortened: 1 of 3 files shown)


compiler/transforms/schedule_memref_linalg.py
```diff
@@ -0,0 +1,302 @@
...
+    For `dim` = 2, will return:
+        (d2) -> d2
+    """
+    return canonicalize_map(
+        map.replace_dims_and_symbols(
+            [AffineConstantExpr(0) for _ in range(dim)]
+            + [AffineDimExpr(i) for i in range(map.num_dims - dim)],
+            [],
+            map.num_dims - dim,
+            0,
```
JosseVanDelm, line 34: Does this mean we should we also add todo's here for adding dynamic shapes?

compiler/transforms/schedule_memref_linalg.py
```diff
@@ -0,0 +1,302 @@
...
+    op_type_rewrite_pattern,
+)
+from xdsl.rewriter import InsertPoint
+
+from compiler.util.canonicalize_affine import canonicalize_map
+
+
+def disable_dims(map: AffineMap, dim: int) -> AffineMap:
+    """
+    Returns an affine map with the first `dim` dimensions set to 0
```
JosseVanDelm, line 19: wondering whether we should replace `first` with `leftmost`

compiler/transforms/schedule_memref_linalg.py
```diff
@@ -0,0 +1,302 @@
...
+    """
+    bounds = bounds.copy()
+    bounds.insert(0, bounds.pop(dim - 1))
+    return bounds
+
+
+def split_dim(map: AffineMap, dim: int, template_bound: int) -> AffineMap:
+    """
+    Returns the bounds and a new affine map with the `dim` dimension split up into two
+    This translates to creating two for loops with adjuste bounds from one for loop
```
JosseVanDelm, line 72:
```suggestion
    This translates to creating two for loops with adjusted bounds from one for loop
```

compiler/transforms/schedule_memref_linalg.py
```diff
@@ -0,0 +1,302 @@
...
+def rotate_bounds(bounds: list[int | None], dim: int) -> list[int | None]:
+    """
+    Returns the bounds after rotating dims, as in rotate_dims
+    """
+    bounds = bounds.copy()
+    bounds.insert(0, bounds.pop(dim - 1))
+    return bounds
+
+
+def split_dim(map: AffineMap, dim: int, template_bound: int) -> AffineMap:
```
JosseVanDelm, line 69: Your guide refers to this as tiling, but e.g. halide calls 2D loop splitting tiling, and 1D loop splitting just splitting.
Maybe change `tiling` `->` `splitting` there?

compiler/transforms/schedule_memref_linalg.py
```diff
@@ -0,0 +1,302 @@
...
+    def match_and_rewrite(self, op: memref_stream.GenericOp, rewriter: PatternRewriter):
+        if any(isinstance(operand.type, stream.StreamType) for operand in op.operands):
+            # Already streamified
+            return
+
+        if not (
+            isinstance(op.library_call, builtin.StringAttr)
+            and op.library_call.data in ("snax_alu", "snax_gemm")
+        ):
+            raise NotImplementedError("panic!")
```
JosseVanDelm, line 128:
```suggestion
            supported = ("snax_alu", "snax_gemm")
            isinstance(op.library_call, builtin.StringAttr)
            and op.library_call.data in supported
        ):
            raise NotImplementedError(f"Currently only the following library calls are supported: {" ,".join(supported)}")
```

compiler/transforms/schedule_memref_linalg.py
```diff
@@ -0,0 +1,302 @@
...
+                    break
+
+                # else rotate the for loops
+                schedule = tuple(rotate_dims(map, schedule_dim + 1) for map in schedule)
+                schedule_bounds = rotate_bounds(schedule_bounds, schedule_dim + 1)
+
+            if not match:
+                raise RuntimeError("failed to match template and schedule")
+
+            # now, check bounds and design potential transoformation map
```
JosseVanDelm, line 184:
```suggestion
            # now, check bounds and design potential transformation map
```

compiler/transforms/schedule_memref_linalg.py
```diff
@@ -0,0 +1,302 @@
...
+                schedule_bounds = split_bounds(
+                    schedule_bounds, schedule_dim, template_bound
+                )
+
+        # if we get here we have now successfully determined a schedule for the operator
+        # this can be implemented with a memref streaming region
+        # the rest of this function just creates the wrapping memref streaming region
+        # the original generic op stays the same
+        # the indexing maps of that generic op are now pretty meaningless, but should
+        # not be looked at in following passes i think
```
JosseVanDelm, line 214: Can you instead put this "wrapping" inside a separate function?

tests/filecheck/transforms/schedule-memref-linalg.mlir
JosseVanDelm, line 1: Can you also add the test for the gemm (for now?) ?

JosseVanDelm (commented):
Hi @jorendumoulin this is a super impressive PR!

It is so well-structured that I feel a bit bad giving suggestions/criticism :)

I'm not approving now, because I first want your opinion on these issues:
* It feels a bit icky to have separated functions working on bounds and maps, can we not generate a new class that keeps them together, which would only require one function implementation of e.g. `rotate`?
* Please add a test for `snax_gemm`
* Consider splitting the internal structure of the rewrite_pattern in two different functions, one that does the scheduling, and one that does the wrapping in the streaming region, I think that would improve reading quality, and it would also probably increase the clarity of possible error messages.

compiler/transforms/schedule_memref_linalg.py
jorendumoulin: yes 😢

compiler/transforms/schedule_memref_linalg.py
jorendumoulin: seems more clear yes

compiler/transforms/schedule_memref_linalg.py
jorendumoulin: yes!

tests/filecheck/transforms/schedule-memref-linalg.mlir
jorendumoulin: yeppers

jorendumoulin:
> Hi @jorendumoulin this is a super impressive PR!
> 
> It is so well-structured that I feel a bit bad giving suggestions/criticism :)
> 
> I'm not approving now, because I first want your opinion on these issues:
> 
>     * It feels a bit icky to have separated functions working on bounds and maps, can we not generate a new class that keeps them together, which would only require one function implementation of e.g. `rotate`?

Yes I agree, I will do that. However, my downstream dev branches have changed some of this functionality so to avoid having resolve all of the conflicts, I will do this after they stabilize if that's allright :)

> 
>     * Please add a test for `snax_gemm`
> 
>     * Consider splitting the internal structure of the rewrite_pattern in two different functions, one that does the scheduling, and one that does the wrapping in the streaming region, I think that would improve reading quality, and it would also probably increase the clarity of possible error messages.

I can definitely separate the scheduling and wrapping, but same comment as before, to avoid breaking my dev branches, I'd like to have this as a todo for later :)

JosseVanDelm (approved):

PR #206, opened 2024-08-26
transform: (stream-snaxify) accelerator attribute hoisting

jorendumoulin:
When transforming the `memref_stream.streaming_region_op` to `snax_stream`, we must add an accelerator attribute to the `snax_stream` op, which is not included in `memref_stream`. For this, we just look at the generic op in the body and see if it has a library call assigned to it

### DIFF
--- a/compiler/transforms/stream_snaxify.py
+++ b/compiler/transforms/stream_snaxify.py
@@ -43,7 +43,7 @@ class HoistAcceleratorAttribute(RewritePattern):
 
             if accelerator and inner_op.library_call != accelerator:
                 raise RuntimeError(
-                    "multiple different accelerator dispatches found in streaming region op"
+                    "multiple different accelerator dispatches found in a single memref_stream.streaming_region op"
                 )
 
             accelerator = inner_op.library_call.data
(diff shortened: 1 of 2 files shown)


compiler/transforms/stream_snaxify.py
```diff
@@ -17,6 +18,40 @@
...
+            # look for memref_stream.generic with library call
+            if not isinstance(inner_op, memref_stream.GenericOp):
+                continue
+
+            if not inner_op.library_call:
+                continue
+
+            if accelerator and inner_op.library_call != accelerator:
+                raise RuntimeError(
+                    "multiple different accelerator dispatches found in streaming region op"
```
JosseVanDelm, line 46:
```suggestion
                    "multiple different accelerator dispatches found in single memref_stream.streaming_region op"
```
Can you add a test to check if this error is thrown?

JosseVanDelm (changes requested):
Okay! Can you add two small filecheck tests? One test to check whether the error is thrown and one to check whether it works?

compiler/transforms/stream_snaxify.py
jorendumoulin: This seems to be a bit uncomfortable. xdsl testing system does not really allow for checking runtime errors in filechecks. Specific exceptions are made for parsing errors and verify errors, but this is neither. In all other passes, when runtime errors are thrown, they aren't really checked 😬 (for example: insert-accfg-op)

JosseVanDelm: oh, allright then :)

JosseVanDelm (approved):
Thanks sir!

compiler/transforms/stream_snaxify.py
jorendumoulin: Yes!

PR #207, opened 2024-08-26
Add double buffering

JonasCrols:
This PR adds a pass that automatically performs double buffering on all loops following the pattern 'Copy In -> Compute -. Copy Out'. It supports both even and uneven amounts of iteration. Currently, nested loops are not supported


jorendumoulin:
replaced by pipeline logic

PR #211, opened 2024-08-28
add streamer-gemm accfg kernel

jorendumoulin:
This PR implements the streamer-gemm through accfg with the streamer interface.
A new kernel has been added with the stream lowering.
For dynamic sizes, the lowering with the C function call is still used.

The old `simple_matmul` tests are now completely removed, as the `streamer_matmul` do everything they did, and more, with a better system.


kernels/streamer_matmul/matmul_stream.mlir
```diff
@@ -1,4 +1,5 @@
-func.func @simple_matmul(%arg0: memref<16x16xi8>, %arg1: memref<16x16xi8, strided<[1, 16], offset:0>>, %arg2: memref<16x16xi32>) {
+// For constant parameters, the stream lowering is enabled
```
JosseVanDelm, line 1:
```suggestion
// For static shapes, the stream lowering is enabled
```

JosseVanDelm: Maybe move this comment to the makefile instead of this file

JosseVanDelm (approved):
:fire: :fire: :fire: :fire_engine: :fire_engine: :fire_engine:

PR #212, opened 2024-08-28
 transform: (stream_snaxify) extend to all data layouts and more dimensions

jorendumoulin:
This allows _any_ layout that has an affine map defined to be used by the stream lowering.
There are also some fixes to include more dimensions for the gemm, these are maybe somewhat unethical.

### DIFF
--- a/compiler/transforms/stream_snaxify.py
+++ b/compiler/transforms/stream_snaxify.py
@@ -128,7 +128,6 @@ class MemrefStreamToSnaxPattern(RewritePattern):
                     "Access patterns with symbols are not supported yet."
                 )
 
-            streamer_config.data.temporal_dim()
             spat_dim = streamer_config.data.spatial_dim()
 
             # extremely dirty fix:
--- a/compiler/transforms/stream_snaxify.py
+++ b/compiler/transforms/stream_snaxify.py
@@ -132,6 +132,8 @@ class MemrefStreamToSnaxPattern(RewritePattern):
             spat_dim = streamer_config.data.spatial_dim()
 
             # extremely dirty fix:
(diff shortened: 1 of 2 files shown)


compiler/transforms/stream_snaxify.py
```diff
@@ -130,24 +128,34 @@ def generate_one_list(n: int, i: int):
...
                 )
 
-            temp_dim = streamer_config.data.temporal_dim()
+            streamer_config.data.temporal_dim()
             spat_dim = streamer_config.data.spatial_dim()
 
+            # extremely dirty fix:
+            if spat_dim == 2:
+                spat_dim = 3
+
```
jorendumoulin, line 137: This is the largest problem here, the access pattern of the matmul has 3 spatial for loops, but the gemm only has 2 spatial streamers, because on for loop is always irrelevant. I have no fix for this yet, just an idea how to fix it, with "virtual streamers". This only works because the gemm is the only one with a spat_dim of 2

compiler/transforms/stream_snaxify.py
```diff
@@ -130,24 +128,36 @@ def generate_one_list(n: int, i: int):
                     "Access patterns with symbols are not supported yet."
                 )
 
-            temp_dim = streamer_config.data.temporal_dim()
+            streamer_config.data.temporal_dim()
```
JosseVanDelm, line 131: This value is not used?

JosseVanDelm (changes requested):
Impressive that you can do this with just 24 lines of code :smile: can you also add a test for TSLs though?

JosseVanDelm (approved):
Yes thank you!

PR #215, opened 2024-08-28
Update to new mlir

JonasCrols:
Update mlir to new commit, same commit as xDSL uses


JonasCrols:
This PR is now a draft as first a new docker container is needed containing the same mlir version xDSL is on

jorendumoulin:
resolved in #240

PR #216, opened 2024-08-28
Add ZigZag Tiling passes

JonasCrols:
This PR adds a sequence of passes used to automatically tile MatMuls into the optimal tiling pattern found using ZigZag.


JonasCrols:
This PR is now a draft as first both the add-c Gemm and the new mlir version (same version as xDSL) need to be added

jorendumoulin:
this work has been integrated into praxis

PR #217, opened 2024-08-29
transforms: (set-memory-layout) make tsls more static

jorendumoulin:
The tsl's were given dynamic strides and bounds, even for static problem sizes.
The assumption that everything downstream easily handles dynamic layouts is not always as convenient :)
This makes sure the layout attribute is exactly as statically determined as the memref it is assigned to.

### DIFF
--- a/compiler/ir/tsl/tiled_strided_layout.py
+++ b/compiler/ir/tsl/tiled_strided_layout.py
@@ -162,7 +162,9 @@ class TiledStridedLayout:
             next_stride = next(
                 (
                     (dim, depth, stride_self)
-                    for dim, depth, stride_self in self_strides
+                    for dim, depth, stride_self in sorted(
+                        self_strides, key=lambda x: x[2].bound is None
+                    )
                     if stride_self.step == current_stride
                 ),
                 None,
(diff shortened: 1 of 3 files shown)


compiler/transforms/set_memory_layout.py
```diff
@@ -34,12 +35,39 @@ def match_and_rewrite(self, linalg_op: linalg.Generic, rewriter: PatternRewriter
             library_call = linalg_op.library_call.data
 
         # check for library call
-        if library_call == "snax_gemm":
+        if library_call == "snax_gemm" or library_call == "snax_gemm_stream":
+            # the layout should be as static as the memref is. no more, no less
+            # get i, j, k
```
JosseVanDelm, line 40: ?
```suggestion
            # get m, n, k
```

compiler/ir/tsl/tiled_strided_layout.py
```diff
@@ -162,7 +162,9 @@ def largest_common_contiguous_block(
             next_stride = next(
                 (
                     (dim, depth, stride_self)
-                    for dim, depth, stride_self in self_strides
+                    for dim, depth, stride_self in sorted(
+                        self_strides, key=lambda x: x[2].bound is None
+                    )
```
JosseVanDelm, line 167: This looks a bit cryptic, what is happening here? Can you add a comment?

JosseVanDelm (approved):
Okay!

PR #218, opened 2024-08-30
Switch region dispatch to use core index directly

JosseVanDelm:
This makes it easier to use the function constant pinning pass of Anton

### DIFF
--- a/compiler/transforms/dispatch_regions.py
+++ b/compiler/transforms/dispatch_regions.py
@@ -1,4 +1,5 @@
 from collections.abc import Callable, Iterable
+from dataclasses import dataclass
 
 from xdsl.context import MLContext
 from xdsl.dialects import arith, builtin, func, scf
@@ -16,7 +17,10 @@ from xdsl.traits import SymbolTable
 from compiler.util.dispatching_rules import dispatch_to_compute, dispatch_to_dm
 
 
+@dataclass
 class DispatchRegionsRewriter(RewritePattern):
+    nb_cores: int
+
     @op_type_rewrite_pattern
     def match_and_rewrite(self, func_op: func.FuncOp, rewriter: PatternRewriter):
         def dispatcher(block: Block, core_cond: builtin.i1, dispatch_rule: Callable):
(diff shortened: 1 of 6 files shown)


JosseVanDelm:
I'm silently waiting for all lit tests to fail

runtime/include/snax_rt.h
```diff
@@ -68,6 +68,4 @@ void _mlir_ciface_snax_dma_2d_transfer(size_t *source, size_t *destination,
   snrt_dma_wait_all();
 }
 
-int _mlir_ciface_snax_is_dm_core() { return snrt_is_dm_core(); }
-
-int _mlir_ciface_snax_is_compute_core() { return snrt_is_compute_core(); }
+uint32_t _mlir_ciface_snax_global_core_idx() { return snrt_global_core_idx(); }
```
jorendumoulin, line 71: is it not better to use cluster_core_idx?

compiler/transforms/dispatch_regions.py
```diff
@@ -60,46 +60,58 @@ def dispatcher(block: Block, core_cond: builtin.i1, dispatch_rule: Callable):
 
         ## dispatch dm core ops, insert function call
         # in dominator block if changes made
-        func_call_dm = func.Call("snax_is_dm_core", [], [builtin.i1])
+
+        # FIXME: currently assuming that DM core is @ index 1 and compute @ index 0
```
jorendumoulin, line 64: unfortunately, this will break my mlperf flow already, maybe for now, an argument for this pass specifying the number of cores in this cluster (defaulting to 2) would be nice? :)

jorendumoulin: compute core can be left at 0 for now, but dm core must be `nb_cores-1`

jorendumoulin (approved):

compiler/transforms/dispatch_regions.py
JosseVanDelm: Oh okay!

JosseVanDelm: It's there for you to @jorendumoulin !

PR #219, opened 2024-08-31
Pin code regions to cores

JosseVanDelm:
Uses constant function pinning to remove all if statement clutter from dispatch_regions. Awaits #222


tests/filecheck/transforms/snax-pin-core-end-to-end.mlir
```diff
@@ -0,0 +1,84 @@
+// RUN: ./compiler/snax-opt %s -p dispatch-regions{nb_cores=3\ pin_to_constants=true},function-constant-pinning,snax-to-func | mlir-opt-17 --canonicalize --inline| filecheck %s
```
JosseVanDelm, line 1: I guess I should remove the --inline, since @AntonLydike specifically outlined the function calls because I requested to do so :( . @AntonLydike can we make the outlining optional at some point do you think?

compiler/transforms/dispatch_regions.py
```diff
@@ -20,6 +20,7 @@
 @dataclass
 class DispatchRegionsRewriter(RewritePattern):
     nb_cores: int
+    pin_to_constants: bool
```
jorendumoulin, line 23: please add docstring on what pin_to_constants means

compiler/transforms/dispatch_regions.py
```diff
@@ -66,8 +67,20 @@ def dispatcher(block: Block, core_cond: builtin.i1, dispatch_rule: Callable):
         # in dominator block if changes made
 
         # FIXME: currently assuming that DM core is nb_cores - 1 and compute @ index 0
+
+        func_call = func.Call("snax_cluster_core_idx", [], [builtin.i32])
+
+        if self.pin_to_constants:
```
jorendumoulin, line 73: I would remove the `pin_to_constants` flag alltogether, as it is a little bit confusing - why would we want it to be false?
If we want to debug, we can just disable the `pin-to-functions` pass? Then it is just a harmless unused attribute.
If you still want to keep it, maybe just let it default to true.

compiler/transforms/dispatch_regions.py
```diff
@@ -128,10 +141,12 @@ class DispatchRegions(ModulePass):
     name = "dispatch-regions"
 
     nb_cores: int = 2  # amount of cores
+    pin_to_constants: bool = False  # add constant pinning to function call
```
jorendumoulin, line 144:
```suggestion
    pin_to_constants: bool = False  # add constant pinning attributes to function call
```

jorendumoulin (approved):

compiler/transforms/dispatch_regions.py
JosseVanDelm: Okay actually I agree! Thanks for the feedback Joren!

PR #224, opened 2024-09-03
[WIP] Some Changes I forgot to Upstream

AntonLydike:

PR #227, opened 2024-09-03
Add fix in scoped_setups

JosseVanDelm:
Fix in scoped_setups, originally in #224 , but now cherry-picked from 9060797ccb8ef1ebe67078382ba1c60a5f5d569c

### DIFF
--- a/compiler/transforms/accfg_config_overlap.py
+++ b/compiler/transforms/accfg_config_overlap.py
@@ -10,10 +10,9 @@ from xdsl.pattern_rewriter import (
     op_type_rewrite_pattern,
 )
 from xdsl.rewriter import InsertPoint
-from xdsl.traits import is_side_effect_free
 
 from compiler.dialects import accfg
-from compiler.inference.helpers import iter_ops_range, previous_ops_of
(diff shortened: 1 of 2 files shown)


compiler/transforms/accfg_config_overlap.py
```diff
@@ -171,6 +166,10 @@ def match_and_rewrite(self, op: accfg.SetupOp, rewriter: PatternRewriter, /):
             op.in_state.index - 1
         )  # -1 because the first block arg is the loop index
 
+        # also, if any operation between us and the loop start is not side effect free, abort
+        if any(not is_side_effect_free(prev_op) for prev_op in previous_ops_of(op)):
+            return
+
```
JosseVanDelm, line 172: So we can remove this now @AntonLydike

AntonLydike: Yes, it appears that we reached the conclusion that this was unneeded

AntonLydike:
:muscle:

AntonLydike (approved):
:muscle:

PR #231, opened 2024-09-06
accelerators: (snax streamers) add streamer flags

jorendumoulin:
This PR extends the streamer configuration attributes so that each streamer dimension can have a flag assigned to it.
This allows us to express the irrelevance of certain dimensions or possible reuse within the accelerator, such that we can program the streamers correctly.

Additionally, this pr also deprecates and starts moving away from the global `spatial_dim` and `temporal_dim` functions. These are only valid when the temporal/spatial dim is equal for every operand, which is only the case for very simple examples.

### DIFF
--- a/compiler/accelerators/streamers/streamers.py
+++ b/compiler/accelerators/streamers/streamers.py
@@ -35,7 +35,7 @@ class StreamerFlag(StrEnum):
 
     def __bool__(self) -> bool:
         """
-        Overrides the default boolean conversion to ensure that the 'None'
+        Overrides the default boolean conversion to ensure that the 'Normal'
         flag evaluates to False and all other flags evaluate to true
         """
         return self is not StreamerFlag.Normal
--- a/compiler/accelerators/streamers/streamers.py
+++ b/compiler/accelerators/streamers/streamers.py
@@ -16,7 +16,7 @@ class StreamerFlag(StrEnum):
 
     Attributes:
     -----------
-    - None: 'n'
+    - Normal: 'n'
       Indicates that no special flags apply.
     - Irellevant : 'i'
       Indicates a dimension is irrelevant for this operand.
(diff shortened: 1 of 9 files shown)


compiler/transforms/stream_snaxify.py
```diff
@@ -130,12 +130,6 @@ def generate_one_list(n: int, i: int):
 
             spat_dim = streamer_config.data.spatial_dim()
 
-            # extremely dirty fix:
-            # FIXME: this only works because gemm is the only one with
-            # two spat_dims. This must be fixed with some "virtual" spatial dim
-            if spat_dim == 2:
-                spat_dim = 3
-
```
jorendumoulin, line 138: Solving this is the main takeaway for this PR. If a spatial dimension is irrelevant to a certain operand, the streamer does not exist and should not be programmed. With the streamer flags this is now properly handled.

compiler/accelerators/streamers/streamers.py
```diff
@@ -8,19 +10,83 @@ class StreamerType(StrEnum):
...
+            (StreamerFlag.Normal,) * spatial_dim,
+        )
+
+    @property
+    def temporal_dim(self):
+        return len(self.temporal_dims)
+
+    @property
+    def spatial_dim(self):
+        return len(self.spatial_dims)
```
jorendumoulin, line 89: The temporal and spatial dim for the streamers of a given operand is still valid.

compiler/accelerators/streamers/streamers.py
```diff
@@ -42,6 +108,7 @@ def size(self) -> int:
         """
         return len(self.streamers)
 
+    @deprecated("Please do not use this function, it is only valid in trivial cases")
     def temporal_dim(self) -> int:
         """
```
jorendumoulin, line 113: These functions try to determine a single temporal dim for the streamers of several operands, but can only be done if they are all equal. This is often not the case.

jorendumoulin: However, my mlperf dev branch still uses this, so I am not removing it just yet to make merging them more graceful.

JosseVanDelm (approved):
Impressive!

PR #232, opened 2024-09-06
dialects: (snax) refactor tsl to use element offset instead of byte offset

jorendumoulin:


compiler/transforms/stream_snaxify.py
```diff
@@ -110,11 +110,8 @@ def generate_one_list(n: int, i: int):
             # Mapping from data to memory:
             assert isinstance(memref_type := op.operands[operand].type, MemRefType)
 
-            # TODO: fix element offset in tsl to avoid this shit
-            if isinstance(memref_type.layout, TiledStridedLayoutAttr):
-                data_mem_map: AffineMap = memref_type.get_affine_map()
-            else:
-                data_mem_map: AffineMap = memref_type.get_affine_map_in_bytes()
+            # Mapping from data to memory:
+            data_mem_map: AffineMap = memref_type.get_affine_map_in_bytes()
```
jorendumoulin, line 114: this is the main result of this pr

JosseVanDelm (approved):
Nice! Can you maybe also add a python test for the `in_bytes = True` field?

PR #233, opened 2024-09-07
add kernel dialect

jorendumoulin:
start migrating this logic from `util` and formalize them as ops.
Kernel ops can be used to match accelerator templates

### DIFF
--- a/compiler/dialects/kernel.py
+++ b/compiler/dialects/kernel.py
@@ -30,7 +30,9 @@ class QuantizedBinaryOp(BinaryOp):
 @irdl_op_definition
 class MulOp(KernelOp, BinaryOp):
     name = "kernel.mul"
-
+    assembly_format = (
+        "$lhs `,` $rhs attr-dict `:` type($lhs) `,` type($rhs) `->` type($result)"
+    )
 
 @irdl_op_definition
 class AddOp(KernelOp, BinaryOp):
--- a/compiler/dialects/kernel.py
+++ b/compiler/dialects/kernel.py
@@ -41,20 +41,18 @@ class AddOp(KernelOp, BinaryOp):
 @irdl_op_definition
 class MacOp(KernelOp, BinaryOp):
     name = "kernel.mac"
-    acc = operand_def(IntegerType)
     assembly_format = (
-        "$lhs `,` $rhs `acc``:` $acc attr-dict `:`"
-        "type($lhs) `,` type($rhs) `,` type($acc) `->` type($result)"
+        "$lhs `,` $rhs attr-dict `:`"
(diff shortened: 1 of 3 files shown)


tests/filecheck/dialects/kernel/ops.mlir
```diff
@@ -0,0 +1,29 @@
+// RUN: XDSL_ROUNDTRIP
+// RUN: XDSL_GENERIC_ROUNDTRIP
+
+// CHECK: builtin.module {
+// CHECK-GENERIC: "builtin.module"() ({
+
+%0 = "test.op"() : () -> i32
+// CHECK-NEXT:   %0 = "test.op"() : () -> i32
+// CHECK-NEXT-GENERIC:   %0 = "test.op"() : () -> i32
```
AntonLydike, line 9: This should be `CHECK-GENERIC-NEXT:` instead of `CHECK-NEXT-GENERIC`

jorendumoulin: oops! thanks for spotting this!

compiler/dialects/kernel.py
```diff
@@ -0,0 +1,58 @@
...
+        "$lhs `,` $rhs attr-dict `:` type($lhs) `,` type($rhs) `->` type($result)"
+    )
+
+
+@irdl_op_definition
+class MacOp(KernelOp, BinaryOp):
+    name = "kernel.mac"
+    assembly_format = (
+        "$lhs `,` $rhs attr-dict `:`" "type($lhs) `,` type($rhs) `->` type($result)"
+    )
```
JosseVanDelm, line 46: Why is this slightly different from the others?
Also, can we not move the assembly format to either ABCs?

compiler/dialects/kernel.py
```diff
@@ -0,0 +1,58 @@
+from abc import ABC
+
+from xdsl.dialects.builtin import IntegerType
+from xdsl.ir import Dialect
+from xdsl.irdl import operand_def
+from xdsl.irdl.operations import irdl_op_definition, result_def
+from xdsl.parser import IRDLOperation
+
+
```
JosseVanDelm, line 9: Can you add a comment somewhere to state what this dialect's purpose is?

JosseVanDelm (approved):
I like this! It's nice and short -- I think we can even make it shorter!
Maybe at some point we should add a bit more verification to this, but not necessarily at this point :). Thanks!

compiler/dialects/kernel.py
jorendumoulin: apperently moving them into the abc's breaks xdsl, and it is not recognized. something to do with the @irdl_op_definition magic

PR #234, opened 2024-09-09
add convert-linalg-to-kernel pass

jorendumoulin:
This pass allows to lower `linalg.generic` with arith bodies to kernel ops, based on specified patterns.
If a kernel op allows a lowering through this path, it inherits from `Parsable` and specifies a specific parsing region.

### DIFF
--- a/compiler/dialects/kernel.py
+++ b/compiler/dialects/kernel.py
@@ -16,7 +16,6 @@ class KernelOp(IRDLOperation, ABC):
     """
 
     ...
-class KernelOp(IRDLOperation, ABC): ...
 
 
 class Parsable(ABC):
@@ -68,7 +67,7 @@ class AddOp(KernelOp, BinaryOp, Parsable):
 @irdl_op_definition
 class MacOp(KernelOp, BinaryOp, Parsable):
     name = "kernel.mac"
-    assembly_format = "$lhs `,` $rhs attr-dict `:`" "type($lhs) `,` type($rhs) `->` type($result)"
+    assembly_format = "$lhs `,` $rhs attr-dict `:` type($lhs) `,` type($rhs) `->` type($result)"
 
     @property
     def parsing_region(self) -> Region:
--- a/compiler/dialects/kernel.py
+++ b/compiler/dialects/kernel.py
@@ -20,7 +20,8 @@ class KernelOp(IRDLOperation, ABC):
 
 class Parsable(ABC):
     @property
-    def parsing_region(self) -> Region: ...
+    def parsing_region(self) -> Region:
+        ...
 
 
 class BinaryOp:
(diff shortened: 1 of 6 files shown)


compiler/transforms/convert_linalg_to_kernel.py
```diff
@@ -0,0 +1,73 @@
...
+    their body matches a specific kernel op defined in
+    the kernel dialect. Replaces the body with the relevant
+    kernel op if this is true.
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, linalg_op: linalg.Generic, rewriter: PatternRewriter):
+        for op_def in Kernel.operations:
+            if not issubclass(op_def, Parsable):
+                continue
```
JosseVanDelm, line 46: Is there any op we would define here that is not `Parsable`?

compiler/dialects/kernel.py
```diff
@@ -28,36 +36,106 @@ class QuantizedBinaryOp(BinaryOp):
...
+class QMacOp(KernelOp, QuantizedBinaryOp, Parsable):
     name = "kernel.qmac"
     assembly_format = (
         "$lhs `,` $rhs `zp_lhs``:` $zp_lhs `zp_rhs``:` $zp_rhs attr-dict `:`"
         " type($lhs) `,` type($rhs) `,` type($zp_lhs) `,` type($zp_rhs) `->` type($result)"
     )
 
+    @property
+    def parsing_region(self) -> Region:
+        @Builder.implicit_region(
```
JosseVanDelm, line 118: I find this name a bit confusing, because you are not parsing a region here, you are emitting it to be easier to be parsed, later, right?

compiler/transforms/convert_linalg_to_kernel.py
```diff
@@ -0,0 +1,73 @@
...
+            try:
+                kernel_op = op_def(
+                    operands=linalg_op.body.block.args[:-1],
+                    result_types=[linalg_op.body.block.args[-1].type],
+                )
+            except ValueError:
+                # wrong number of operands for this operation, no match
+                continue
```
JosseVanDelm, line 54: This seems like a very cheeky way to check which operation we should pick?
I'd prefer if this were a bit more specific and defined  in a positive way.
Instead of trying out all constructors and using the one that works, let's pick the ones that have the same amount of operands? Do you think it is possible to pick this statically?

compiler/dialects/kernel.py
JosseVanDelm: Can we rename this to `equivalent_region` ?

compiler/transforms/convert_linalg_to_kernel.py
```diff
@@ -0,0 +1,73 @@
...
+    op_type_rewrite_pattern,
+)
+from xdsl.rewriter import InsertPoint
+
+from compiler.dialects.kernel import Kernel, Parsable
+
+
+def check_block(block_a: Block, block_b: Block) -> bool:
+    """
+    Verify if two blocks are equal to each other,
```
JosseVanDelm, line 18:
```suggestion
def check_kernel_equivalence(block_a: Block, block_b: Block) -> bool:
    """
    Verify if two blocks are equivalent to each other,
```

compiler/transforms/convert_linalg_to_kernel.py
JosseVanDelm: What happens if something else goes wrong here? Like a type error?

JosseVanDelm (commented):
I took a while for me to understand what you are trying to do here.
IIUC What you are doing is:
* Get all operands of a linalg generic
* Try all constructors of the operations defined there.
* If a constructor works, check if the region being defined is equivalent.

I do agree that this is:
* More declarative than the previous approach
* Quite short to write

What bothers me:
* We are relying on errors for proper operation of this mechanism. This feels like an anti-pattern

compiler/transforms/convert_linalg_to_kernel.py
jorendumoulin: yes, kernel.rescale from https://github.com/KULeuven-MICAS/snax-mlir/pull/236 for example

compiler/transforms/convert_linalg_to_kernel.py
jorendumoulin: I figured out how to do this check

JosseVanDelm (approved):
Love it! Thanks!

PR #237, opened 2024-09-16
transforms: (dispatch-kernels) move to template system

jorendumoulin:
This PR cleans up the dispatch-kernels pass, to make it a little bit less hardcoded and more reusable.

New way of working:
An accelerator optionally specifies a compute kernel template by inheriting the `DispatchTemplate` class.
This template specifies supported kernels of the `kernel` dialect, and the datatypes it supports for all the operands.
The more generic pass dispatches the kernels based on this template, for the acceleratros that have been inserted into the ir with an accelerator op.

If you don't want to use this system, you can always still specify a custom lowering pattern in `dispatch-kernels`

I removed the `simple_mult` kernel, because it is a bit outdated and does not really fit this model as it has no accfg implementation.


compiler/transforms/dispatch_kernels.py
jorendumoulin, line 1: the diff of this file is messed up

compiler/dialects/kernel.py
```diff
@@ -30,9 +30,12 @@ class BinaryOp:
     result = result_def(IntegerType)
 
 
-class QuantizedBinaryOp(BinaryOp):
+class QuantizedBinaryOp:
+    lhs = operand_def(IntegerType)
+    rhs = operand_def(IntegerType)
     zp_lhs = operand_def(IntegerType)
     zp_rhs = operand_def(IntegerType)
+    result = result_def(IntegerType)
```
jorendumoulin, line 39: this was necessary because otherwise the ordering of operands is wrong, if i used inheritance the ordering was (zp_lhs, zp_rhs, lhs, rhs) and there was seemingly no way to change this

tests/filecheck/transforms/dispatch_kernels.mlir
jorendumoulin, line 1: also a bit of a messed up diff

JosseVanDelm (approved):
Looks very cool! Can you have a look at my comments though? Thanks so much!

compiler/accelerators/dispatching.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@dataclass
+class SupportedKernel:
+    kernel_type: type[KernelOp]
+    operand_types: Iterable[Attribute]
+
+
+@dataclass
+class DispatchTemplate(Accelerator, ABC):
```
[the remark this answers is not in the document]
jorendumoulin: We dispatch using the accelerator name, which is a property of Accelerator.
This makes sure that everything that inherits dispatchtemplate must also inherit accelerator, so the name of a dispatchtemplate is always available.

compiler/transforms/dispatch_kernels.py
```diff
@@ -10,176 +11,101 @@
...
-                imaps_target[i] != op.indexing_maps.data[i].data
-                for i in range(len(imaps_target))
-            ]
-        ):
+        matched_accelerator: type[DispatchTemplate] | None = None
+
+        for accelerator in self.accelerators:
+            if matched_accelerator:
+                # already match found
+                break
```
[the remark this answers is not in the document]
jorendumoulin: the nested for loop breaks out of the loop on a match (line 74), but only breaks the inner for loop, not the outer one, so this check is required

PR #238, opened 2024-09-17
add debugging dialect and pass

jorendumoulin:
This PR adds some testing infrastructure to better inspect larger kernels such as neural networks.
It allows for the automatic insertion of debug statements before and after every linalg.generic kernel.
The debug statements are later transformed into function calls. 
The function definition can be provided by the user in the `main.c` file, where you can inspect the operands given to you by the debug statements.

Because this is not really core infrastructure, i put all the files in a `test/` subfolder

### DIFF
--- a/compiler/dialects/test/debug.py
+++ b/compiler/dialects/test/debug.py
@@ -1,0 +1,58 @@
+from __future__ import annotations
+
+from xdsl.dialects.builtin import (
+    MemRefType,
+    StringAttr,
+)
+from xdsl.ir import Attribute, Dialect, SSAValue
+from xdsl.irdl import (
+    IRDLOperation,
+    irdl_op_definition,
+    operand_def,
+    prop_def,
+)
+
+
+@irdl_op_definition
+class DebugOp(IRDLOperation):
+    """
+    Run a debug statement, passing operands to a C function
+    to allow inspection of the memref operands.
+    """
+
+    name = "debug.debug"
+
+    op_a = operand_def(MemRefType[Attribute])
+    op_b = operand_def(MemRefType[Attribute])
+    op_c = operand_def(MemRefType[Attribute])
+
+    # what linalg kernel am i debugging?
+    debug_type = prop_def(StringAttr)
+
+    # am i debuggin before or after the op?
+    when = prop_def(StringAttr)
+
+    # at what memory level is the data i am inspecting?
+    level = prop_def(StringAttr)
+
+    def __init__(
+        self,
+        op_a: SSAValue,
+        op_b: SSAValue,
+        op_c: SSAValue,
+        debug_type: str,
+        when: str,
+        level: str = "L1",
+    ):
+        super().__init__(
+            operands=[op_a, op_b, op_c],
+            result_types=[],
+            properties={
+                "debug_type": StringAttr(debug_type),
+                "when": StringAttr(when),
+                "level": StringAttr(level),
+            },
+        )
+
+
+Debug = Dialect("debug", [DebugOp])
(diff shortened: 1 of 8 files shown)


compiler/transforms/test/debug_to_func.py
```diff
@@ -0,0 +1,71 @@
...
+            f"debug_{op.debug_type.data}", [ptr_a, ptr_b, ptr_c, when], []
+        )
+        ops_to_insert.append(func_call)
+        rewriter.replace_matched_op(ops_to_insert)
+
+        func_decl = func.FuncOp.external(
+            f"debug_{op.debug_type.data}",
+            [builtin.i32, builtin.i32, builtin.i32, builtin.i32],
+            [],
+        )
```
JosseVanDelm, line 58: Are these runtime thingies also added to this PR?

compiler/dialects/test/debug.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@irdl_op_definition
+class DebugOp(IRDLOperation):
+    """
+    Run a debug statement, passing operands to a C function
+    to allow inspection of the memref operands.
+    """
+
+    name = "debug.debug"
```
JosseVanDelm, line 23: Nit: Can I would already pick a more specific name here if I'm honest.

JosseVanDelm (commented):
Hi! This is very nice.

There are a  few related efforts that you should know about/look into (maybe :) ):
@AntonLydike made a `dump` dialect once for debugging MPI ( I believe?)
Me and @AntonLydike once made a `printf` dialect. I think it is still available in xDSL!

Apart from that. No need to put it in a separate test folder in my opinion. I think debug infrastructure is important, and justified as core infra.

compiler/transforms/test/debug_to_func.py
jorendumoulin: no, the idea is you can specify your own in `main.c` and inspect anything you want in C code

jorendumoulin:
I looked into printf, but is limited to printing ints and chars it seems, and might be difficult to extend. This approach is just very convenient because you can write custom logic in C to debug the data your linalg is operating on. I think this is still more convenient to quickly change what you want to debug. Writing C sucks, but writing MLIR sucks even harder

compiler/transforms/test/debug_to_func.py
JosseVanDelm: Oh okay cool! Can we at least have one test for this mechanism?

jorendumoulin: The streamer-alu kernel uses this mechanism, and outputs the nice code:
```
Debugging at t = 0 with A at 80002F28, B at 80002FA8, C at 80003028
i0 -> A=44, B=39, C=0
i1 -> A=47, B=87, C=0
i2 -> A=64, B=46, C=0
i3 -> A=67, B=88, C=0
i4 -> A=67, B=81, C=0
Debugging at t = 1 with A at 10000040, B at 10000100, C at 100001C0
i0 -> A=44, B=39, C=0
i1 -> A=47, B=87, C=0
i2 -> A=64, B=46, C=0
i3 -> A=67, B=88, C=0
i4 -> A=67, B=81, C=0
Debugging at t = 2 with A at 10000040, B at 10000100, C at 100001C0
i0 -> A=44, B=39, C=83
i1 -> A=47, B=87, C=134
i2 -> A=64, B=46, C=110
i3 -> A=67, B=88, C=155
i4 -> A=67, B=81, C=148
Debugging at t = 3 with A at 80002F28, B at 80002FA8, C at 80003028
i0 -> A=44, B=39, C=83
i1 -> A=47, B=87, C=134
i2 -> A=64, B=46, C=110
i3 -> A=67, B=88, C=155
i4 -> A=67, B=81, C=148
```

jorendumoulin: this prints before copying in L3, after copying in L1, after compute in L1 and then after copying back to L3

PR #240, opened 2024-09-17
bump snax to v0.1.13

jorendumoulin:
most notable change is a more recent version of mlir included in the docker container


JosseVanDelm (approved):
:superhero:

PR #241, opened 2024-09-18
remove kernel type util

jorendumoulin:
Everything has migrated to the kernel dialect, so this can be removed


JosseVanDelm (approved):
:star_struck:

PR #242, opened 2024-09-18
transform: (convert-tosa-to-kernel) use correct output type

jorendumoulin:
The output type was not set correctly in the resulting linalg.generic. The outs of a linalg on tensors must have the correct type, as it uses the `DestinationStyleOpInterface`. This output type can be generated with a `tensor.empty` operation

### DIFF
--- a/compiler/transforms/convert_tosa_to_kernel.py
+++ b/compiler/transforms/convert_tosa_to_kernel.py
@@ -1,8 +1,8 @@
 from xdsl.builder import Builder
 from xdsl.context import MLContext
-from xdsl.dialects import builtin, linalg, tosa
+from xdsl.dialects import arith, builtin, linalg, tensor, tosa
 from xdsl.dialects.builtin import IntegerAttr, i1, i8, i32
-from xdsl.ir import BlockArgument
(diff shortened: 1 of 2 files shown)


compiler/transforms/convert_tosa_to_kernel.py
```diff
@@ -45,24 +46,13 @@ def match_and_rewrite(self, rescale_op: tosa.RescaleOp, rewriter: PatternRewrite
...
-
-        max_int = clamp_op.max_int.value.data
-        assert_int8(max_int)
-
-        min_int = clamp_op.min_int.value.data
-        assert_int8(min_int)
-
+        shift = assert_int8(rescale_op.shift.data.data[0].data)
+        max_int = assert_int8(clamp_op.max_int.value.data)
+        min_int = assert_int8(clamp_op.min_int.value.data)
```
jorendumoulin, line 55: this makes pyright happy, to make sure that the values are integers

JosseVanDelm: Nice!

JosseVanDelm (approved):
Ok nice!

PR #243, opened 2024-09-19
dialects: (kernel) turn rescale properties into attributes

jorendumoulin:
A weird bug in MLIR causes the properties of my unregistered dialect to be removed during the fold tensor cast pattern when applying canonicalization.
Running this through canonicalize shows the bug: https://gist.github.com/jorendumoulin/03b55e6b467914de74a010e60590acb8

For now, this turns the properties into attributes such that we are not bothered by this bug


JosseVanDelm (approved):
Okay, if you have reported this behaviour, can you link that report here?

PR #244, opened 2024-09-19
transform: (preprocess-mlperftiny) initialize preprocessing pass

jorendumoulin:
This PR introduces the preprocessing pass for mlperftiny networks.
It is a collection of lowering patterns and ad-hoc transformations to get the network into the nice format to be ingested by the compiler.
Perhaps with a better frontend for neural networks, this would be slightly less ugly, but this performs quite nicely for now.

This awaits #243, who's changes are already present here.

### DIFF
--- a/compiler/transforms/convert_tosa_to_kernel.py
+++ b/compiler/transforms/convert_tosa_to_kernel.py
@@ -25,6 +25,8 @@ def assert_int8(val) -> int:
 class RescaleClampPattern(RewritePattern):
     """
     Transform rescale clamp into a kernel.rescale op
+    If there is no clamp op after the rescaling op,
+    the clamping defaults to the range of the int8 output type (-128, 127)
     """
 
     @op_type_rewrite_pattern
--- a/compiler/transforms/frontend/preprocess_mlperf_tiny.py
+++ b/compiler/transforms/frontend/preprocess_mlperf_tiny.py
@@ -26,7 +26,7 @@ class InsertStaticFunctionCall(RewritePattern):
     def match_and_rewrite(self, op: builtin.ModuleOp, rewriter: PatternRewriter):
         """
         Insert a function with static arguments them to dynamic ones,
-        calls the existing main function, and then casts the dyanmic output
+        calls the existing main function, and then casts the dynamic output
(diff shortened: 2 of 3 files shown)


compiler/transforms/frontend/preprocess_mlperf_tiny.py
```diff
@@ -0,0 +1,312 @@
...
+        if not isinstance((const_op := const_input.op), arith.Constant):
+            return
+        if not isinstance((intattr := const_op.value), builtin.IntegerAttr):
+            return
+        if intattr.value.data != 0:
+            return
+
+        # erase the op
+        rewriter._replace_all_uses_with(op.results[0], op.outputs[0])
+        rewriter.erase_matched_op()
```
JosseVanDelm, line 140: I guess the difference w.r.t `InsertMemoryClears` is that this has to do with initialization, whereas that has to do more with freeing/allocation?

JosseVanDelm (approved):
Nice work Joren!

compiler/transforms/convert_tosa_to_kernel.py
```diff
@@ -51,8 +52,12 @@ def match_and_rewrite(self, rescale_op: tosa.RescaleOp, rewriter: PatternRewrite
         multiplier = rescale_op.multiplier.data.data[0].data
         assert isinstance(multiplier, int)
         shift = assert_int8(rescale_op.shift.data.data[0].data)
-        max_int = assert_int8(clamp_op.max_int.value.data)
-        min_int = assert_int8(clamp_op.min_int.value.data)
+        if isinstance(clamp_op, tosa.ClampOp):
+            max_int = assert_int8(clamp_op.max_int.value.data)
+            min_int = assert_int8(clamp_op.min_int.value.data)
+        else:
```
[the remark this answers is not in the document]
jorendumoulin: clamping is performed either by rescaleop or by  clampop. if it is done by clampop, we take those values, if it is done by rescaleop, it is done with -128 and 127

compiler/transforms/frontend/preprocess_mlperf_tiny.py
jorendumoulin: Yes!

PR #246, opened 2024-09-19
streamer_config: update to the latest streamer features

jorendumoulin:
new features include:

- sometimes the bounds are programmed separately, sometimes not 😫 (this should be resolved after the streamer merge but I want to first support neural networks before we use the new streamers)
- readerwriter is now also a possiblity!
- readertranspose is now also a possibility!

### DIFF
--- a/compiler/accelerators/snax.py
+++ b/compiler/accelerators/snax.py
@@ -113,12 +113,16 @@ class SNAXStreamer(ABC):
 
         # loop bound registers
         if not self.streamer_config.data.separate_bounds:
-            loop_bounds: Sequence[IntAttr] = op.stride_patterns.data[0].upper_bounds.data
+            loop_bounds: Sequence[IntAttr] = op.stride_patterns.data[
+                0
+            ].upper_bounds.data
         else:
             loop_bounds: Sequence[IntAttr] = []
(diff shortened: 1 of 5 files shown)


compiler/accelerators/snax.py
```diff
@@ -112,16 +112,21 @@ def _generate_streamer_setup_vals(
...
-                    cst.result,
-                )
-                for loop_bound in loop_bounds
-            ]
-        )
+        for operand, streamer in enumerate(self.streamer_config.data.streamers):
+            upper_bounds = op.stride_patterns.data[operand].upper_bounds.data
+            while streamer.temporal_dim > len(upper_bounds):
+                # if not all temporal bounds are used, insert 1's
+                upper_bounds = (IntAttr(1),) + upper_bounds
```
JosseVanDelm, line 119: I would write this like you are writing it in the comment (if possible), not sure if the if statement is necessary anymore then:
```suggestion
           # pad unused temporal bounds with 1's
           upper_bounds = (IntAttr(1),) * (streamer.temporal_dim - len(upper_bounds))  + upper_bounds
```

JosseVanDelm (approved):
Cool!

PR #248, opened 2024-09-19
kernels: (streamer-matmul) use gemmx

jorendumoulin:
this pr replaces the streamer-gemm with the new gemmx accelerator for the kernels:

- dynamic tests are removed as they are not yet supported by the stream lowering
- half_tiled matmul is removed as the transform_matmul also covers this case


compiler/accelerators/snax_gemmx.py
```diff
@@ -150,4 +151,54 @@ def _generate_setup_vals(
...
+            csr1 = c0
+            csr2 = c0
+
+            # get zero points for gemm
+            assert isinstance(qmac.zp_lhs, BlockArgument)
+            zp_a = generic_op.inputs[qmac.zp_lhs.index]
+            assert isinstance(qmac.zp_rhs, BlockArgument)
+            zp_b = generic_op.inputs[qmac.zp_rhs.index]
+
+            # bitwise and with 8b'11111111 to don't fuck up bitpacking in case of negative values
```
JosseVanDelm, line 182:
```suggestion
            # bitwise and with 8b'11111111 to prevent extraneous sign extension
```

JosseVanDelm: I don't understand why this is necessary actually. You can perform explicit unsigned extension of 8-bit values with extui?

compiler/transforms/schedule_memref_linalg.py
```diff
@@ -141,6 +142,25 @@ def match_and_rewrite(self, op: memref_stream.GenericOp, rewriter: PatternRewrit
...
+                ]
+                template_bounds = (None, None, None, 8, 8, 8)
+            else:
+                # rescale function of gemmx
+                M, K, m, k = (AffineDimExpr(i) for i in range(4))
+                template = [
+                    AffineMap(4, 0, (M * 8 + m, K * 8 + k)),
+                    AffineMap(4, 0, (M * 8 + m, K * 8 + k)),
+                ]
+                template_bounds = (None, None, 8, 8)
```
JosseVanDelm, line 163: Could you already write them to be movable?

compiler/transforms/stream_snaxify.py
```diff
@@ -155,13 +156,52 @@ def generate_one_list(n: int, i: int):
...
+                        upper_bounds=snax_stride_patterns[3].upper_bounds,
+                        temporal_strides=[0] * 3,
+                        spatial_strides=[0, 4, 32],
+                    ),
+                )
+
+                # point C to zero allocated row in TCDM.
+                new_inputs.append(
+                    arith.Constant.from_int_and_width(0x1000_0040, builtin.IndexType())
+                )
```
JosseVanDelm, line 199: I'm wondering whether we should ask for hardware supported zero memory

JosseVanDelm (commented):
Is dynamic shape support for streamer lowerings planned/in the works? I agree with moving forward, but this feels (maybe wrongly) like a reduction of our test surface area?

compiler/accelerators/snax_gemmx.py
jorendumoulin: The problem is that the values are in 32 bit format, after default linalg lowerings, so extui is not possible

compiler/transforms/schedule_memref_linalg.py
jorendumoulin: What does this mean?

compiler/transforms/stream_snaxify.py
jorendumoulin: New streamers can be masked such that they output zero values, so we could use that.

jorendumoulin:
> Is dynamic shape support for streamer lowerings planned/in the works? I agree with moving forward, but this feels (maybe wrongly) like a reduction of our test surface area?

Indeed, this removes a test, but for a lowering I feel we don't want to support. I don't immediately plan on dynamic lowerings for streamers, but can do it when necessary

compiler/transforms/schedule_memref_linalg.py
JosseVanDelm: oeps

compiler/accelerators/snax_gemmx.py
JosseVanDelm: Right. Okay :) . Still would like the comment to be a bit more descriptive :)

JosseVanDelm (approved):

PR #249, opened 2024-09-20
Move benchmark utilities to util folder

JosseVanDelm:
Since these are actually utilties, we move these to the util folder


jorendumoulin (approved):

util/gendata.py
jorendumoulin, line 1: what does this mean?

JosseVanDelm: For some reason this part of the library was adapted to have run permissions (`chmod +x`, even though it is not an executable, so I `chmod -x`'ed it :smile: )

PR #251, opened 2024-09-20
kernels: (rescale) add kernel

jorendumoulin:
This PR adds lowering capabilites for the rescale simd function of `snax_gemmx`, and a kernel to test it out

### DIFF
--- a/compiler/accelerators/snax.py
+++ b/compiler/accelerators/snax.py
@@ -137,23 +137,29 @@ class SNAXStreamer(ABC):
                 (IntAttr(0),) * (streamer.temporal_dim - len(temporal_strides))
             ) + temporal_strides
             for dim, flag in enumerate(streamer.temporal_dims):
-                stride = temporal_strides[dim].data
+                stride = temporal_strides[dim]
                 if flag == StreamerFlag.Irrelevant:
                     # Irrelevant temporal strides should be zero
                     assert stride == 0
-                cst = arith.Constant.from_int_and_width(stride, i32)
+                cst = arith.Constant.from_int_and_width(stride.data, i32)
                 result.append(([cst], cst.result))
 
         # spatial strides
         for operand, streamer in enumerate(self.streamer_config.data.streamers):
+            strides = iter(op.stride_patterns.data[operand].spatial_strides.data)
+            stride : IntAttr | None = next(strides, None)
             for dim, flag in enumerate(streamer.spatial_dims):
-                stride = op.stride_patterns.data[operand].spatial_strides.data[dim].data
+                # stride = op.stride_patterns.data[operand].spatial_strides.data[dim].data
                 if flag == StreamerFlag.Irrelevant:
                     # Irrelevant spatial strides are not programmed
-                    assert stride == 0
+                    if stride and stride.data == 0:
+                        stride = next(strides)
                     continue
+                assert stride # stride should be intattr here
                 cst = arith.Constant.from_int_and_width(stride, i32)
                 result.append(([cst], cst.result))
+                stride = next(strides, None)
+            assert not stride # all strides must be processed by this point
 
         # input & output base pointers
         result.extend(([], x) for x in op.inputs)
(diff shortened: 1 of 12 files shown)


compiler/accelerators/snax.py
jorendumoulin, line 1: this is just some cleaning up

compiler/accelerators/snax_gemmx.py
```diff
@@ -189,17 +190,59 @@ def _generate_setup_vals(
...
+            csr0 = ops_to_add[-1].results[0].op.results[0]
+            ops_to_add.extend(pack_bitlist([double_round, min_int], [8, 0]))
+            csr1 = ops_to_add[-1].results[0].op.results[0]
+            csr2 = mult.result
+
+            loop_bound = prod(x.data for x in op.stride_patterns.data[0].upper_bounds)
+            loop_bound = arith.Constant.from_int_and_width(loop_bound, i32)
+            ops_to_add.append(loop_bound)
+
+        else:
```
jorendumoulin, line 236: this extracts and computes the correct values for csr's based on the `kernel.rescale` op

compiler/transforms/set_memory_layout.py
jorendumoulin, line 1: this pass should be next on the schedule to fix in a better way.
For now this needs to assign a certain layout for simd because the inputs are connected to the wide tcdm

compiler/transforms/set_memory_layout.py
```diff
@@ -9,11 +9,98 @@
     op_type_rewrite_pattern,
 )
 
+from compiler.dialects.kernel import QMacOp, RescaleOp
 from compiler.dialects.snax import LayoutCast
 from compiler.dialects.tsl import TiledStridedLayoutAttr
 from compiler.ir.tsl import Stride, TiledStride, TiledStridedLayout
 
 
+class AddMemoryLayoutSIMD(RewritePattern):
```
JosseVanDelm, line 18: Add a docstring here please that clarifies that it only applies this in case of a SIMD operation

compiler/accelerators/snax_gemmx.py
JosseVanDelm: Nice ! Add as a comment in the code :)

compiler/transforms/stream_snaxify.py
```diff
@@ -125,25 +126,54 @@ def generate_one_list(n: int, i: int):
                     "Access patterns with symbols are not supported yet."
                 )
 
-            spat_dim = streamer_config.data.spatial_dim()
-
-            temporal_strides = []
-            spatial_strides = []
-            upper_bounds = []
+            # Get the streamer
+            # FIXME: some hardcoded fix because there is no system yet to map a specific operand
```
JosseVanDelm, line 130: Yeah this is becoming a bit much :) but okay to "fix" in the future.

JosseVanDelm (approved):
Thanks! This looks good, modulo a bit of comments for clarity! :fist_right: :fist_left:

compiler/accelerators/snax.py
```diff
@@ -137,23 +137,29 @@ def _generate_streamer_setup_vals(
...
             for dim, flag in enumerate(streamer.spatial_dims):
-                stride = op.stride_patterns.data[operand].spatial_strides.data[dim].data
+                # stride = op.stride_patterns.data[operand].spatial_strides.data[dim].data
                 if flag == StreamerFlag.Irrelevant:
                     # Irrelevant spatial strides are not programmed
-                    assert stride == 0
+                    if stride and stride.data == 0:
+                        stride = next(strides, None)
                     continue
+                assert stride  # stride should be intattr here
```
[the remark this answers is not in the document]
jorendumoulin: as stride is IntAttr | None as two lines above, this case is handled

compiler/accelerators/snax.py
```diff
@@ -137,23 +137,29 @@ def _generate_streamer_setup_vals(
...
                 if flag == StreamerFlag.Irrelevant:
                     # Irrelevant temporal strides should be zero
                     assert stride == 0
-                cst = arith.Constant.from_int_and_width(stride, i32)
+                cst = arith.Constant.from_int_and_width(stride.data, i32)
                 result.append(([cst], cst.result))
 
         # spatial strides
         for operand, streamer in enumerate(self.streamer_config.data.streamers):
+            strides = iter(op.stride_patterns.data[operand].spatial_strides.data)
```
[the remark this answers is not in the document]
jorendumoulin: unfortunately this does not work in conjunction with the `dim, flag` for loop

PR #253, opened 2024-09-23
transforms (convert-kernel-to-linalg) init pass

jorendumoulin:
In case you converted to `kernel` dialect but decided not to dispatch anything, this pass converts the linalg-parsable ops back to core dialects such that llvm understands them.


JosseVanDelm (approved):
Nice!

PR #254, opened 2024-09-23
Add SNAX specific events to the tracing script

AntonLydike:
Adds SNAX specific event tracer to the trace_to_perfetto utilty:

The following three events are discerned to produce those traces:

* a "pre-launch" event, that:
        - starts at the first CSR write to a config register
        - ends after all launch fields have been written to
        - records the number of setup instructions executed
 * a "launched but not waiting" event, that:
        - starts immediately after the "pre-launch" event
        - ends on the second "write 0 to launch addr"
 * a "waiting" event, that:
        - starts on the first read from the barrier address
        - ends when that read return 1
        
This should be general enough to support any SNAX accelerator, although I (@JosseVanDelm) am not entirely sure how to verify that in CI.

The utility was already useable standalone, but this patch makes it more difficult to specify a bunch of options that are invalid.
* You can now use `--sequential` to run the tracer in sequential mode, which is useful for debugging with pdb.
* You can use `--accelerator` to specify a SNAX accelerator in order to activate the SNAX specific event tracer (this functionality is also now available for `process_traces` in `SNAXBenchmark`.
* Kernel annotation is now optional  with `--annotate-kernels` and the tool will warn you if you do not specify the `--elf`-file and the `--addr2line` executable

Important note:
In perfetto you might want to change your time format to be anything but timecode, because otherwise it seems to mess up the numbers (not sure why, and also not sure how we can change that default behaviour)
image

### DIFF
--- a/util/tracing/snax_event_generator.py
+++ b/util/tracing/snax_event_generator.py
@@ -13,7 +13,7 @@ class _CurrentEvent:
 
 @dataclass
 class SettingUp(_CurrentEvent):
-    number_setups: int = field(default=1)
+    number_of_setups: int = field(default=1)
     number_of_launches: int = field(default=0)
     """
     Since we only end this region when we see the *second* write to launch, we need to keep track of the number
@@ -27,7 +27,7 @@ class Launched(_CurrentEvent):
     """
     How many zero writes to the launch address have happened
     """
-    number_setups: int = field(default=0)
+    number_of_setups: int = field(default=0)
 
 
 @dataclass
@@ -85,14 +85,14 @@ class SNAXAcceleratorEventGenerator(EventGenerator):
         elif isinstance(self.state, SettingUp):
             # if it's a normal setup, count a setup ins
             if ins.csr in self.fields:
-                self.state.number_setups += 1
+                self.state.number_of_setups += 1
             # if it's a launch insn
             elif ins.csr in self.launch_fields:
                 # check that we haven't met the launch threshold yet
                 if self.state.number_of_launches < len(self.launch_fields):
                     self.state.number_of_launches += 1
                     # this launch counts as a setup
-                    # self.state.number_setups += 1
+                    # self.state.number_of_setups += 1
                 # otherwise, end this event, switch to launch event
                 else:
                     events.append(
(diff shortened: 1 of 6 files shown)


JosseVanDelm:
@AntonLydike I rebased this, found a few errors, and added some functionality

util/tracing/snax_event_generator.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def cycle(self, state: TraceState) -> list[DurationEvent]:
+        ins = state.instruction
+        if not isinstance(ins, CSRInstruction):
+            return []
+        events: list[DurationEvent] = []
+
+        # if state is None, check if we are setting up, if so, switch state to "setting up"
+        if self.state is None:
+            if (ins.csr in self.fields) or (ins.csr in self.launch_fields):
```
JosseVanDelm, line 80: This was a bit wrong before :smile: :
```
            if ins.csr in self.fields or self.launch_fields):
```
Which is almost always true :sweat_smile:

JosseVanDelm:
@AntonLydike I can not request a review from you, but I'd appreciate it if you could have a look at this :)

JosseVanDelm:
@jorendumoulin and @AntonLydike, can we get this in?
Once we move to snakemake we probably will want to make independent parts a bit more independent and more configurable, but for now this should already be useful

util/tracing/trace_to_perfetto.py
```diff
@@ -41,13 +52,35 @@ def parse_arguments():
...
+    snax_accelerators = []
+    for accelerator, acc_class in AcceleratorRegistry().registered_accelerators.items():
+        if issubclass(acc_class, SNAXAccelerator):
+            snax_accelerators.append(accelerator)
+
+    parser.add_argument(
+        "--accelerator",
+        choices=snax_accelerators,
+        default=None,
+        help="SNAX accelerator for SNAX Event Annotator",
```
jorendumoulin, line 81: does this actually do something?

jorendumoulin (approved):
Nice!

util/tracing/trace_to_perfetto.py
JosseVanDelm: This code makes sure that you can only call the argumentparser with SNAX-like accelerators

JosseVanDelm:
Good to go modulo #298

PR #257, opened 2024-09-24
transforms (test-add-mcycle-around-launch) init pass

jorendumoulin:
To measure the raw execution time of accelerators in benchmarks


JosseVanDelm (approved):
Beautiful! It collides a bit with the goal of #254 , but I think this is useful still!

PR #258, opened 2024-09-24
bump snax to v0.2.1

jorendumoulin:
new streamers!
~~new bugs!~~

The most drastic change are in the spatial streamers, which are now all 1-dimensional. We need a better system to support code-generation for this, for which I have some ideas. In this PR, the spatial strides are hardcoded to [8]. Because in all of our tests the data accessed spatially is contiguous, this works, but is off course not a generally accepted solution.

I propose to do this in a follow up-PR to keep these changes nice and separated.

The changes in this PR mostly relate to the order in which the CSRs come, which has changed somewhat.
additionally, the `separate_bounds` flag is removed as it is now always separate
the readerwriter streamertype is removed as it is probably just a hardware concept that is better modeled as a separate reader and writer from a compiler point of view

### DIFF
--- a/compiler/accelerators/snax.py
+++ b/compiler/accelerators/snax.py
@@ -119,17 +119,10 @@ class SNAXStreamer(ABC):
             )
 
             # spatial strides
-            spatial_strides = []
             for dim, flag in enumerate(streamer.spatial_dims):
                 stride = op.stride_patterns.data[operand].spatial_strides.data[dim].data
-                if flag == StreamerFlag.Irrelevant:
-                    # Irrelevant spatial strides are not programmed
-                    assert stride == 0
-                    continue
                 cst = arith.Constant.from_int_and_width(stride, i32)
-                spatial_strides.append(([cst], cst.result))
-            # innermost spatial stride is not programmed
-            result.extend(spatial_strides[1:])
+                result.append(([cst], cst.result))
 
             # loop bounds
             upper_bounds = op.stride_patterns.data[operand].upper_bounds.data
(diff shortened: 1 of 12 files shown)


jorendumoulin:
oh no failing tests are giving green signals :'(

compiler/transforms/stream_snaxify.py
```diff
@@ -127,20 +126,20 @@ def generate_one_list(n: int, i: int):
...
+            template_bounds: tuple[int | None, ...] = ()
+            if acc_str == "snax_alu":
+                template_bounds = (None, 4)
+            elif acc_str == "snax_gemm":
+                template_bounds = (None, None, None, 8, 8, 8)
+            elif acc_str == "snax_gemmx":
+                if len(op.inputs) > 1:
+                    # gemm
+                    template_bounds = (None, None, None, 8, 8, 8)
                 else:
```
JosseVanDelm, line 141:
```suggestion
                # simd only
                else:
```

compiler/accelerators/snax.py
```diff
@@ -144,96 +152,38 @@ def _generate_streamer_setup_vals(
...
-                c0 = arith.Constant.from_int_and_width(0, i32)
-                result.append(([c0], c0.result))
-                # in the current version of the streamer,
-                # transpose fields are bit packed together,
-                # so only one field for all transpose ops
-                # this should be fixed in the new streamer
-                break
+                # if we want to disable transpose, we obviously need to program
+                # a 1 to the transpose field (much logical)
+                # threrefore always program to 1
```
JosseVanDelm, line 160: Although I share the sentiment and appreciate the joke, I'm not a big fan of sarcastic comments in code. Let's keep it very clear for other people as well :)

JosseVanDelm (approved):
Nice work!

PR #259, opened 2024-09-24
kernels: return 1 on error

jorendumoulin:
Right now, the kernels return the number of errors in `main.c`. However, this seems to cause some kernels succeeding in CI even though they are wrong. For example, upon returning 256, CI will not fail.

Examples:

Succeeding test when returning 256: https://github.com/KULeuven-MICAS/snax-mlir/actions/runs/11012247027/job/30578039634
Failing test when returning 1: https://github.com/KULeuven-MICAS/snax-mlir/actions/runs/11012262776/job/30578088724

This PR just makes sure all kernels return 1 if an error occurs

### DIFF
--- a/kernels/simple_copy/main.c
+++ b/kernels/simple_copy/main.c
@@ -47,5 +47,5 @@ int main() {
       nerr += 1;
   }
 
-  return err != 0;
+  return nerr != 0;
 }
--- a/kernels/simple_copy/main.c
+++ b/kernels/simple_copy/main.c
@@ -47,5 +47,7 @@ int main() {
       nerr += 1;
   }
 
-  return nerr;
+  if (nerr > 0)
+    return 1;
+  return 0;
 }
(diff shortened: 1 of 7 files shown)


kernels/simple_copy/main.c
```diff
@@ -47,5 +47,7 @@ int main() {
       nerr += 1;
   }
 
-  return nerr;
+  if (nerr > 0)
+    return 1;
+  return 0;
```
JosseVanDelm, line 52: If you want, you can do this with a oneliner
```suggestion
  return nerr != 0;
```

JosseVanDelm (approved):
Nice! But you seem to have forgotten a few kernels?
how about alloc, rescale, ...?
Can you also update the benchmarks then?

jorendumoulin:
> Nice! But you seem to have forgotten a few kernels? how about alloc, rescale, ...? Can you also update the benchmarks then?

sure! I just took the list of kernels from ci, rescale did not exist back then, but alloc seems to not be included in ci lol?

jorendumoulin:
and the alloc kernel does not even work :'(, I'll make an issue for this one

PR #263, opened 2024-09-30
introduce streamer opts

jorendumoulin:
a better way to specify the custom options of streamers, every streamer now has a set of "options" to determine its capabilities

### DIFF
--- a/compiler/accelerators/snax.py
+++ b/compiler/accelerators/snax.py
@@ -9,7 +9,7 @@ from xdsl.ir import Operation, SSAValue
 
 from compiler.accelerators.accelerator import Accelerator
 from compiler.accelerators.streamers import StreamerConfiguration
-from compiler.accelerators.streamers.streamers import StreamerFlag, StreamerType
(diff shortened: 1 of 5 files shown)


compiler/accelerators/snax.py
```diff
@@ -152,9 +152,15 @@ def _generate_streamer_setup_vals(
                 cst = arith.Constant.from_int_and_width(stride.data, i32)
                 result.append(([cst], cst.result))
 
+            # channel mask option
+            if StreamerOpts.HasChannelMask in streamer.opts:
+                # default to 0 for now
+                c0 = arith.Constant.from_int_and_width(0, i32)
+                result.append(([c0], c0.result))
+
```
JosseVanDelm: Can you also update the SNAX version for this feature in this PR?

JosseVanDelm (approved):
I like this!

compiler/accelerators/snax.py
jorendumoulin: that happens in #264 as there are a lot of unrelated changes which I wanted to separate

PR #264, opened 2024-09-30
bump snax to v0.2.2

jorendumoulin:
This mostly includes the channelwise quantization features of the simd in gemmx!
The lowerings still assume one single quantization factor, thus the same one is programmed to all separate channels


JosseVanDelm (approved):
# Cool

PR #268, opened 2024-10-01
add benchmark outputs

jorendumoulin:


jorendumoulin:
this now uploads artifacts instead of using lfs

jorendumoulin:
i only went back to 1 .gitignore

JosseVanDelm (approved):
Cool!

PR #269, opened 2024-10-02
init gh-pages workflows

jorendumoulin:
While most of the docs can remain on the `gh-pages` branch, the workflow needs to be in main to be able to be manually dispatched or through the cron schedule.

I also think it makes sense to keep the non-auto-generated docs in the main branch as well

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,0 +1,57 @@
+# Sample workflow for building and deploying a Jekyll site to GitHub Pages
+name: Build and deploy gh-pages
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["gh-pages"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+  schedule:
+    - cron: "0 0 * * *" # run daily at 12 am
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,7 +1,5 @@
 # snax stuff
-*generated*
 *logs*
-*results*
```
JosseVanDelm, line 3: Why do these need to go?

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,57 @@
...
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["gh-pages"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+  schedule:
+    - cron: "0 0 * * *" # run daily at 12 am
```
JosseVanDelm, line 13:
```suggestion
  schedule:
    - cron: "0 0 * * 1-5" # run daily at 12 am, but not on weekends
```
:wink:

JosseVanDelm (approved):
Nice!

.gitignore
jorendumoulin: the action needs to commit files that are called result, and this was not happening, pr #268 puts it on the benchmark folder specifically so the convenience of a clean staged file list still remains

.github/workflows/gh-pages.yml
jorendumoulin: github actions does not get time off 😠

PR #272, opened 2024-10-02
transforms: (set-memory-layout) add layout switch

jorendumoulin:
add an argument to the pass to select between banked and cyclic layouts for gemm, to compare the results of both
update the dense benchmark to sweep over this new degree of freedom accordingly

this also defaults to cyclic layout, as it should be the better one

### DIFF
--- a/benchmarks/dense_matmul/genbenchmark.py
+++ b/benchmarks/dense_matmul/genbenchmark.py
@@ -71,7 +71,7 @@ def generate_dense_benchmark(m, n, k) -> SNAXBenchmark:
     write_module_to_file(module, "generated.mlir")
     binary = "generated.x"
     bm = SNAXBenchmark(
-        kernel=f"dense_matmul_generated_{k}x{n}x{m}",
+        kernel=f"dense_matmul_{layout}_{k}x{n}x{m}",
         binary=binary,
         src_dir=str(pathlib.Path.cwd()),
         export_dir=str(pathlib.Path.cwd()),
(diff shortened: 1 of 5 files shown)


tests/filecheck/transforms/set-memory-layout.mlir
```diff
@@ -14,6 +14,6 @@ builtin.module {
 }
 
 
-//CHECK:       %2 = "snax.layout_cast"(%arg0) : (memref<?x128xi8, 1 : i32>) -> memref<?x128xi8, #tsl.tsl<[?, 8] -> (4096, 8), [16, 8] -> (256, 1)>, 1 : i32>
-//CHECK-NEXT:  %3 = "snax.layout_cast"(%arg1) : (memref<128x128xi8, 1 : i32>) -> memref<128x128xi8, #tsl.tsl<[16, 8] -> (256, 1), [16, 8] -> (4096, 8)>, 1 : i32>
+//CHECK:       %2 = "snax.layout_cast"(%arg0) : (memref<?x128xi8, 1 : i32>) -> memref<?x128xi8, #tsl.tsl<[?, 8] -> (1024, 8), [16, 8] -> (64, 1)>, 1 : i32>
+//CHECK-NEXT:  %3 = "snax.layout_cast"(%arg1) : (memref<128x128xi8, 1 : i32>) -> memref<128x128xi8, #tsl.tsl<[16, 8] -> (64, 1), [16, 8] -> (1024, 8)>, 1 : i32>
```
JosseVanDelm, line 18: Can you add a check-prefix for either layout maybe?

compiler/transforms/set_memory_layout.py
```diff
@@ -121,6 +126,8 @@ def match_and_rewrite(self, linalg_op: linalg.Generic, rewriter: PatternRewriter
         else:
             library_call = linalg_op.library_call.data
 
+        assert self.gemm_layout in ("cyclic", "banked")
+
```
JosseVanDelm, line 130: Would it make sense to make this a StrEnum instead?

benchmarks/dense_matmul/genbenchmark.py
```diff
@@ -80,23 +80,31 @@ def generate_dense_benchmark(m, n, k) -> SNAXBenchmark:
...
-        result += f"| {output_report[benchmark]['ideal']} "
-        result += f"| {output_report[benchmark]['utilization']} | \n"
-        avg_utilization += output_report[benchmark]["utilization"]
-        avg_n += 1
-    result += "| average | | |"
-    result += f"{avg_utilization/avg_n} |\n"
+    for layout in ("cyclic", "banked"):
+        result += f"Results for a {layout} layout \n\n"
+        result += "| benchmark | layout | M | N | K | cycles | ideal | utilization |\n"
+        result += "| --- | --- | --- | --- |\n"
```
JosseVanDelm, line 90: I'd like to deduplicate this test a bit from the memory layout. (E.g. have a test that tests utilization vs matrix size.
And have a separate test that tests utilization vs size vs layout ?) But I'm not sure that makes sense?

JosseVanDelm (approved):
Nice! Small comments

tests/filecheck/transforms/set-memory-layout.mlir
jorendumoulin: ok!

compiler/transforms/set_memory_layout.py
jorendumoulin: ok!

benchmarks/dense_matmul/genbenchmark.py
jorendumoulin: it seems now we have everything? (utilization vs matrix size vs layout)? In the outputs, the results are separated by layout

PR #273, opened 2024-10-03
streamers: use channel mask for zero patterns

jorendumoulin:
If we generate a zero pattern with the streamers, we can use the optional channel mask to make sure we do not burden the tcdm with zero requests.

### DIFF
--- a/compiler/accelerators/snax.py
+++ b/compiler/accelerators/snax.py
@@ -5,13 +5,15 @@ from collections.abc import Sequence
 from xdsl.dialects import arith, builtin, llvm
 from xdsl.dialects.builtin import IntAttr, i32
 from xdsl.dialects.scf import Condition, While, Yield
-from xdsl.ir import Operation, SSAValue
+from xdsl.ir import Operation, OpResult, SSAValue
 
 from compiler.accelerators.accelerator import Accelerator
 from compiler.accelerators.streamers import StreamerConfiguration
 from compiler.accelerators.streamers.streamers import StreamerFlag, StreamerOpts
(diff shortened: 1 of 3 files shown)


compiler/accelerators/snax.py
```diff
@@ -112,8 +116,20 @@ def _generate_streamer_setup_vals(
         result: Sequence[tuple[Sequence[Operation], SSAValue]] = []
 
         for operand, streamer in enumerate(self.streamer_config.data.streamers):
+            # streamer must generate zero pattern if it is set to c0
```
JosseVanDelm, line 119:
```suggestion
            # streamer must generate zero pattern if the stream is coming from c0
```

compiler/transforms/stream_snaxify.py
```diff
@@ -223,7 +223,8 @@ def generate_one_list(n: int, i: int):
 
                 # point C to zero allocated row in TCDM.
```
JosseVanDelm, line 224: is this comment still valid?

JosseVanDelm (approved):
Very nice!

compiler/accelerators/snax.py
```diff
@@ -90,6 +92,8 @@ class SNAXStreamer(ABC):
     streamer_setup_fields: Sequence[str]
     streamer_launch_fields: Sequence[str]
 
+    zero_address = 0x1000_0040
+
```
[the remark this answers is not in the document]
jorendumoulin: unfortunately it must be se tot this value (first row of tcdm which is kept free for zeros)

PR #274, opened 2024-10-08
add custom stream dialect

jorendumoulin:
There are certain things a bit annoying about the upstream `stream` + `memref_stream` dialect for our purposes, but I don't know what an optimal solution is, to experiment with this, this creates a more simplified, principled approach to the stream dialect to experiment what would be good for our purposes.

Changes made here:

- Inherent compatability with tensors for easier tiling/splitting/fusing
- Stream types have value semantics, e.g, no `stream.writeable` or `stream.readable`, just `stream.stream`.
- A stream.generic can **only** operate on streams, not on memrefs, and thus is always part of a streaming region
- no attempt to support imperfectly nested loops. to do this, add multiple stream.generics in a stream.streaming region
- stream.generic is kept as a subset of linalg.generic, without specified indexing patterns / iterator types, as they have no relevance on streams
- The access patterns are not specified as custom StridePatterns, but with regular affine maps. These do not include the bounds of the operation, but should always be able to be extracted from the entire op through `get_static_bounds`.
- All of the access patterns should have the same number of dimensions.

### DIFF
--- a/compiler/dialects/stream.py
+++ b/compiler/dialects/stream.py
@@ -31,6 +31,14 @@ from xdsl.irdl import (
     var_result_def,
 )
 
+"""
+Custom `stream` dialect, to simplify things in a more principled approach, including:
+- inherent support for tensors
+- streams with value semantics
+- no specified static bounds in access patterns: they are just affine maps
+- no stream ops outside of streaming regions allowed
+"""
+
 _StreamTypeElement = TypeVar("_StreamTypeElement", bound=Attribute)
 
 
@@ -42,7 +50,7 @@ class StreamType(
     ContainerType[_StreamTypeElement],
 ):
     """
-    A stream type with value sementics.
+    A stream type with value semantics.
     A stream is defined by an element type, and is produced as the result of
     an operation, or through a streaming region op.
     Streams can only be read from, there is no distinction between readable/writeable streams.
@@ -114,14 +122,14 @@ class YieldOp(AbstractYieldOperation[Attribute]):
 @irdl_op_definition
 class GenericOp(IRDLOperation):
     """
-    Generic that operates on streams.
-    As indexing maps / iterators are thus not relevant, they are removed.
+    Generic that operates on streams, similar to a linalg.generic.
+    Indexing maps / iterators are not relevant, so they are not included.
     """
 
     name = "stream.generic"
 
-    inputs = var_operand_def()
-    outputs = var_result_def()
+    inputs = var_operand_def(StreamType)
+    outputs = var_result_def(StreamType)
 
     body = region_def()
 
@@ -159,3 +167,4 @@ Stream = Dialect(
         StreamType,
     ],
 )
+
--- a/compiler/dialects/stream.py
+++ b/compiler/dialects/stream.py
@@ -122,8 +122,6 @@ class GenericOp(IRDLOperation):
     doc = opt_prop_def(StringAttr)
     library_call = opt_prop_def(StringAttr)
 
-    irdl_options = [AttrSizedOperandSegments(as_property=True)]
-
     def __init__(
         self,
         inputs: Sequence[SSAValue],
--- a/compiler/dialects/stream.py
+++ b/compiler/dialects/stream.py
@@ -111,13 +111,13 @@ class YieldOp(AbstractYieldOperation[Attribute]):
     traits = frozenset([IsTerminator()])
 
 
-
 @irdl_op_definition
 class GenericOp(IRDLOperation):
     """
     Generic that operates on streams.
     As indexing maps / iterators are thus not relevant, they are removed.
     """
+
     name = "stream.generic"
 
     inputs = var_operand_def()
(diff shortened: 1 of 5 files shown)


tests/filecheck/transforms/stream-snaxify.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: ./compiler/snax-opt --split-input-file %s -p insert-accfg-op{accelerator=snax_alu},stream-snaxify | filecheck %s
+// RUN: ./compiler/snax-opt --split-input-file %s -p insert-accfg-op{accelerator=snax_alu},stream-snaxify --disable-verify | filecheck %s
```
jorendumoulin, line 1: There is some upstream xdsl verification check that expects a `stream.readable`, from the upstream xdsl dialect. The stream-snaxify still relies on some upstreamness, but I wanted to change this in a separate PR as it would otherwise be too clogged. This simple addition disables the verification checks, but still includes the filecheck to make sure the final result is correct

compiler/dialects/stream.py
```diff
@@ -0,0 +1,161 @@
...
+
+@irdl_attr_definition
+class StreamType(
+    Generic[_StreamTypeElement],
+    ParametrizedAttribute,
+    TypeAttribute,
+    ContainerType[_StreamTypeElement],
+):
+    """
+    A stream type with value sementics.
```
JosseVanDelm, line 45:
```suggestion
    A stream type with value semantics.
```

compiler/dialects/stream.py
```diff
@@ -0,0 +1,161 @@
...
+    name = "stream.yield"
+
+    traits = frozenset([IsTerminator()])
+
+
+@irdl_op_definition
+class GenericOp(IRDLOperation):
+    """
+    Generic that operates on streams.
+    As indexing maps / iterators are thus not relevant, they are removed.
```
JosseVanDelm, line 118:
```suggestion
    Indexing maps / iterators are not relevant, so they are not included.
```

compiler/dialects/stream.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+
+    name = "stream.generic"
+
+    inputs = var_operand_def()
+    outputs = var_result_def()
```
JosseVanDelm, line 124: I guess we also don't need an extra `outputs` operand here, since these only work on streams? Should you also constrain these to have `StreamType`s only?
```suggestion
    outputs = var_result_def(StreamType())
```

compiler/dialects/stream.py
```diff
@@ -0,0 +1,161 @@
...
+    Generic[_StreamTypeElement],
+    ParametrizedAttribute,
+    TypeAttribute,
+    ContainerType[_StreamTypeElement],
+):
+    """
+    A stream type with value sementics.
+    A stream is defined by an element type, and is produced as the result of
+    an operation, or through a streaming region op.
+    Streams can only be read from, there is no distinction between readable/writeable streams.
```
JosseVanDelm, line 48: Something I find a bit odd is that the stream would throw away all the shape information, but I guess a similar thing happens inside a linalg generic?

Also, I assume a stream can only be read from once, since reads "pop" from the stream?
Should we also enforce single SSA use somewhere? (If so, how?)

compiler/dialects/stream.py
```diff
@@ -0,0 +1,161 @@
...
+    irdl_op_definition,
+    opt_prop_def,
+    prop_def,
+    region_def,
+    var_operand_def,
+    var_result_def,
+)
+
+_StreamTypeElement = TypeVar("_StreamTypeElement", bound=Attribute)
+
```
JosseVanDelm, line 35: I would add a link here to the other stream dialect implementation and also add the motivation (description of your pull request) in here somewhere.

compiler/dialects/stream.py
```diff
@@ -0,0 +1,161 @@
...
+
+@irdl_op_definition
+class StreamingRegionOp(IRDLOperation):
+    """
+    An operation that creates streams from tensors or memrefs, which are only available to
+    read from within the body of the operation.
+
+    Within the loop body, memrefs/tensors that are streamed must not be otherwise accessed
+    via any other access means, including extraction (e.g.: memref.view).
```
JosseVanDelm, line 69: I guess this pertains to my previous comment?

compiler/dialects/stream.py
```diff
@@ -0,0 +1,161 @@
...
+class YieldOp(AbstractYieldOperation[Attribute]):
+    name = "stream.yield"
+
+    traits = frozenset([IsTerminator()])
+
+
+@irdl_op_definition
+class GenericOp(IRDLOperation):
+    """
+    Generic that operates on streams.
```
JosseVanDelm, line 117:
```suggestion
    Generic that operates on streams, similar to a linalg.generic.
```

tests/filecheck/transforms/stream-snaxify.mlir
JosseVanDelm: Yikes? I would assume the verifiers to be part of importing the dialect? Lol?

JosseVanDelm (approved):
Looks good!

PR #275, opened 2024-10-08
add convert-linalg-to-stream pass

jorendumoulin:
With the simplifications of the custom stream dialect, this combines the `guarded-convert-linalg-to-stream`, `convert-linalg-to-stream`, `memref-streamify` into a single pass, with a significant loc reduction.

Awaits #274

### DIFF
--- a/compiler/dialects/stream.py
+++ b/compiler/dialects/stream.py
@@ -128,6 +128,7 @@ class GenericOp(IRDLOperation):
 
     name = "stream.generic"
 
+    # inputs can be streams or integers
     inputs = var_operand_def()
     outputs = var_result_def(StreamType)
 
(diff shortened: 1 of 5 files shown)


compiler/dialects/stream.py
```diff
@@ -128,7 +128,7 @@ class GenericOp(IRDLOperation):
 
     name = "stream.generic"
 
-    inputs = var_operand_def(StreamType)
+    inputs = var_operand_def()
```
jorendumoulin, line 131: generics can also have i32 inputs, but it's rabbit hole in python typing hell to make this work without angering pyright, so just like in the regular linalg.generic there is no constraint

JosseVanDelm: Maybe put this as a comment then :)

JosseVanDelm (approved):
Nice! Why are we not removing the previous passes here then?

jorendumoulin:
> Nice! Why are we not removing the previous passes here then?

#276!

PR #277, opened 2024-10-09
add fuse-streaming-regions pass

jorendumoulin:
A pass to fuse multiple streaming regions if their bodies can be executed by the same accelerator.

This pass fuses single producer-consumer streaming regions if the consumer applies an elementwise function to the tensor, with the assumption that this will always be possible to fuse in the accelerator.

This works on tensor operands, so after this pass there should still be some form of bufferization.

### DIFF
--- a/compiler/transforms/fuse_streaming_regions.py
+++ b/compiler/transforms/fuse_streaming_regions.py
@@ -1,0 +1,154 @@
+from dataclasses import dataclass
+
+from xdsl.context import MLContext
+from xdsl.dialects import builtin
+from xdsl.ir import Block, BlockArgument, Region
+from xdsl.ir.affine import AffineMap
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import (
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+from xdsl.rewriter import InsertPoint
+
+from compiler.dialects import stream
+
+
+@dataclass
+class FuseElementwisePattern(RewritePattern):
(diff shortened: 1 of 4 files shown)


compiler/transforms/fuse_streaming_regions.py
```diff
@@ -0,0 +1,168 @@
...
+        rewriter.erase_matched_op()
+
+
+@dataclass(frozen=True)
+class FuseStreamingRegions(ModulePass):
+    name = "fuse-streaming-regions"
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        PatternRewriteWalker(
+            FuseElementwisePattern(), apply_recursively=False
```
JosseVanDelm, line 167: Why can we not apply this pass recursively, now this will maximally fuse two operations, right? I think we should be able to fuse multiple operations together if we like, no?

tests/filecheck/transforms/fuse-streaming-regions.mlir
JosseVanDelm, line 1: What happens if there are more streaming regions? Will they all be fused together?

JosseVanDelm (approved):
Solid! Could you just add a test to see whether it can recursively fuse the rest of the operations?

compiler/transforms/fuse_streaming_regions.py
jorendumoulin: yes the apply recursively is now gone

tests/filecheck/transforms/fuse-streaming-regions.mlir
jorendumoulin: yeppers

PR #278, opened 2024-10-10
add stream-bufferize pass

jorendumoulin:
This pass partially bufferizes streaming regions, such that --one-shot-bufferize of mlir works!


compiler/transforms/snax_bufferize.py
jorendumoulin, line 1: this is necessary because I have not figured out how to properly call the mlir-opt with the complex argements of --one-shot-bufferize. It's also very ugly if we have to include these arguments everywhere

JosseVanDelm (approved):
really cool!

PR #279, opened 2024-10-10
move template definitions to the accelerators

jorendumoulin:
This is much cleaner and solves a couple of fixme's


JosseVanDelm (approved):
:nail_care: beautiful

PR #280, opened 2024-10-10
Add real-life NN matrix multiplication size benchmarks

xiaoling-yi:
We add some extra benchmarking matrix sizes extracted from real NN's representative and heavy layers. Basically, there are from ["lstm", "mnv2", "resnet18", "vit", 'ibertbase'].

The specific matrix size is obtained from Zigzag with the optimal mapping (with the K maximized to maximize output data reuse, the TCDM size is 128KB).


jorendumoulin:
Perfect! I'll resolve the remaining errors :)

benchmarks/dense_matmul/genbenchmark.py
```diff
@@ -128,10 +135,67 @@ def output_log_benchmark(benchmark_name: str, utilization: dict[str, int]) -> st
 
     sizes = list(itertools.product(selected_dims, repeat=3))
 
+    # some other relevant neural network sizes:
```
JosseVanDelm, line 138: @xiaoling-yi could you maybe add a comment (with a link?) here that says where these numbers come from exactly? I think this would be nice for later reference.
>they are from ["lstm", "mnv2", "resnet18", "vit", 'ibertbase'].

Essentially I don't know what this means, and I'd like a comment in code that explains this rather than a comment in a pull request :smile:

JosseVanDelm (commented):
Thanks for your contribution @xiaoling-yi !
I would like us to point in the code a bit more precisely where these numbers come from though.
Also, would it be possible to maybe group them in some way w.r.t where these numbers come from (in first instance just with comments in the code maybe?)
Thank you!

xiaoling-yi:
Hi @JosseVanDelm , I extracted the matrix size of representative layers from the above mentioned neural networks (explicit im2col if CNN). As these layers are too large to fit in the TCDM, I use Zigzag to do the tiling (with optimal mapping). The presented matrix sizes are small tiles that can already fit in TCDM. I can group them to specific networks but maybe too hard to comment on where exactly they came from. Is it Ok?

Also, we can always find better mapping with larger TCDM sizes/better SNAX-GEMM model inside the Zigzag I think. So these number is only a benchmarking reference.

jorendumoulin (approved):
Yes @xiaoling-yi if you could just group them by network that would be very nice!

PR #281, opened 2024-10-14
move alloc-to-global pass out of frontend

jorendumoulin:
when doing bufferization in snax-opt itself, this one is quite useful. Now only used in mlperf, I also want to use it when using `stream-bufferize` and such


JosseVanDelm (approved):
Yesyesyes

PR #282, opened 2024-10-14
transforms: (set-memory-space) add support for streaming regions

jorendumoulin:
same as linalg.generics, force all memrefs to L1 for these operations

### DIFF
--- a/compiler/transforms/set_memory_space.py
+++ b/compiler/transforms/set_memory_space.py
@@ -193,7 +193,7 @@ class InitLinalgMemorySpace(RewritePattern):
 
 class InitStreamMemorySpace(RewritePattern):
     """
-    Covnert all stream.streaming region ops to only use L1
+    Convert all stream.streaming region ops to only use L1
     """
 
     @op_type_rewrite_pattern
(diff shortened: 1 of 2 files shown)


compiler/transforms/set_memory_space.py
```diff
@@ -188,6 +191,41 @@ def get_cast_op(operand) -> None | memref.MemorySpaceCast:
...
+            ] = memref.MemorySpaceCast.from_type_and_target_space(
+                operand, memreftype, L1
+            )
+        rewriter.insert_op(tuple(memory_cast_ops.values()), InsertPoint.before(op))
+
+        # replace all operands to casted values
+        for i in range(len(op.operands)):
+            if op.operands[i] in operands_to_memory_cast:
+                op.operands[i] = memory_cast_ops[op.operands[i]].dest
+
```
JosseVanDelm, line 227: Can you rewrite the InitLinalgMemorySpace to be the same like this?
Now these do the pretty much the same, but are written quite differently.
The linalg one reuses previous memory space casts and creates a new op, the way you've written this pass seems more efficient, and to me they look like they can share the exact same code?

JosseVanDelm (approved):
Okay, but now we have two very similar passes that could share more code.

compiler/transforms/set_memory_space.py
jorendumoulin: Okay! I combined them into one pattern!

JosseVanDelm: Thank you!

PR #283, opened 2024-10-14
transforms: (set-memory-layout) make compatible with streaming region

jorendumoulin:
the pass was only compatible with linalg.generics, this change updates it to also be compatible with streaming regions

### DIFF
--- a/compiler/transforms/set_memory_layout.py
+++ b/compiler/transforms/set_memory_layout.py
@@ -131,7 +131,7 @@ class AddMemoryLayout(RewritePattern):
     ):
         # check if operation is dispatched via library call, as set by e.g.
         # the dispatch-kernels pass
-        #
+
         if isinstance(op, linalg.Generic):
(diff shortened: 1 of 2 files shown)


compiler/transforms/set_memory_layout.py
```diff
@@ -124,31 +126,48 @@ class AddMemoryLayout(RewritePattern):
...
+    def match_and_rewrite(
+        self, op: linalg.Generic | stream.StreamingRegionOp, rewriter: PatternRewriter
+    ):
         # check if operation is dispatched via library call, as set by e.g.
         # the dispatch-kernels pass
-        if linalg_op.library_call is None:
-            return
-        else:
-            library_call = linalg_op.library_call.data
+        #
```
JosseVanDelm, line 134:
```suggestion
```

JosseVanDelm (approved):

PR #284, opened 2024-10-14
transforms: (realize-memref-casts) make compatible with streaming regions

jorendumoulin:
only change needed is to add support for detecting inputs/outputs of streaming regions

### DIFF
--- a/compiler/transforms/realize_memref_casts.py
+++ b/compiler/transforms/realize_memref_casts.py
@@ -10,6 +10,7 @@ from xdsl.pattern_rewriter import (
     op_type_rewrite_pattern,
 )
 
+from compiler.dialects import stream
 from compiler.dialects.snax import LayoutCast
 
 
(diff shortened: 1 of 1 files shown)


compiler/transforms/realize_memref_casts.py
```diff
@@ -94,11 +95,13 @@ def match_and_rewrite(
                 continue
             # check if input
             is_input = False
-            if not isinstance(use_op, linalg.Generic):
+            if isinstance(use_op, linalg.Generic):
```
JosseVanDelm, line 98:
```suggestion
            if isinstance(use_op, linalg.Generic) or isinstance(use_op, stream.StreamingRegionOp):
```
Maybe?

compiler/transforms/realize_memref_casts.py
```diff
@@ -114,6 +117,8 @@ def match_and_rewrite(
             is_output = False
             if isinstance(use_op, linalg.Generic):
                 is_output = op.results[0] in use_op.outputs
```
JosseVanDelm, line 119:
```suggestion
            if isinstance(use_op, linalg.Generic) or isinstance(use_op, stream.StreamingRegionOp):
                is_output = op.results[0] in use_op.outputs
```

JosseVanDelm (approved):
Looks good! Not sure if the suggestions are too important :)

compiler/transforms/realize_memref_casts.py
jorendumoulin: pyright does not agree

PR #285, opened 2024-10-14
add gemm kernel

jorendumoulin:
This PR adds a gemm kernel (D=AxB+C), where the bias addition gets fused with the matmul in the dispatching to gemmx.
The changes compromise simple additions to make this kernel with 4 operands possible.

This is also the first kernel with the new lowering flow:

linalg (on tensors) -> dispatch & schedule -> streaming region -> fuse streaming regions -> only now bufferize -> ... (same as old lowering flow)

This can be seen for example in the `.mlir` file, where the operation is specified on tensors.


JosseVanDelm (approved):
This is really impressive @jorendumoulin !
Just note that a General GEMM actually performs: C = α AB + β C, so we are "still" not fully covered :)

jorendumoulin:
I will leave that one as an exercise for the reader

PR #286, opened 2024-10-14
add add-c to dense matmul kernel

jorendumoulin:
From the updated docs:

> Testing peak utilization of the `snax-gemmx` accelerator, for different matrix sizes. Measures max performance without considering control overhead. This test generates a lot of results and plots to get deep insight into the performance of the accelerator combined with the TCDM interconnect and data layout strategies.
> 
> The following operations are compared:
> 
> - matmul (D = AxB)
> - gemm (D = AxB)
> 
> The following layout strategies are compared:
> 
> - cyclic
> - banked
> 
> The following sizes are run:
> 
> - every combination of (32, 48, 64)
> - some common sizes for neural networks
> 

This PR adds the gemm to this benchmark (previously only matmul)

### DIFF
--- a/benchmarks/dense_matmul/genbenchmark.py
+++ b/benchmarks/dense_matmul/genbenchmark.py
@@ -4,54 +4,57 @@ import pathlib
 from datetime import datetime
 from io import StringIO
 
-from xdsl.builder import ImplicitBuilder
-from xdsl.dialects import arith, builtin, func, linalg
+from xdsl.builder import Builder
+from xdsl.dialects import arith, builtin, func, linalg, tensor
 from xdsl.dialects.builtin import i8, i32
-from xdsl.ir import Block, Region
+from xdsl.ir import BlockArgument
 from xdsl.printer import Printer
 
 from util.snax_benchmark import SNAXBenchmark
 
 
-def create_matrix_multiply(k, m, n):
+def create_matrix_multiply(m, n, k, add_c: bool = False):
     """
-    Generate IR in the form of:
+    Generate IR for a matmul / gemm in the form of:
     ```
     builtin.module {
-        func.func @streamer_matmul(%arg0 : memref<16x16xi8>, %arg1 : memref<16x16xi8,
-                                strided<[1, 16]>>, %arg2 : memref<16x16xi32>) {
+        func.func @streamer_matmul(
+            %arg_a : tensor<16x16xi8>,
+            %arg_b : tensor<16x16xi32>,
+            %arg_c : tensor<16x16xi32>) -> tensor<16x16xi32>{
         %0 = arith.constant 0 : i32
-        linalg.quantized_matmul ins(%arg0, %arg1, %0, %0 : memref<16x16xi8>,
-                                    memref<16x16xi8, strided<[1, 16]>>, i32, i32)
-                                outs(%arg2 : memref<16x16xi32>)
-        func.return
+        %1 = linalg.quantized_matmul ins(%arg_a, %arg_b, %0, %0 : tensor<16x16xi8>, tensor<16x16xi8>, i32, i32)
+                                outs(%arg2 : tensor<16x16xi32>)
+        (optional) := %2 = linalg.add %1, %arg_c ...
+        func.return %1 / %2
         }
     }
     ```
     """
 
-    def get_2d_memref_type(typ, dim_one, dim_two, transpose=False):
-        layout = (
-            builtin.StridedLayoutAttr([1, dim_one]) if transpose else builtin.NoneAttr()
-        )
-        return builtin.MemRefType(typ, [dim_one, dim_two], layout=layout)
-
-    input_types = [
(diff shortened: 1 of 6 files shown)


benchmarks/dense_matmul/genbenchmark.py
```diff
@@ -84,17 +93,20 @@ def output_log(output_report) -> str:
...
     dt_string = datetime.now().strftime("%d/%m/%Y %H:%M:%S")
     result += f"This test was run at {dt_string}\n\n"
-    for layout in ("cyclic", "banked"):
-        result += f"Results for a {layout} layout \n\n"
-        result += "| benchmark | layout | M | N | K | plots | cycles | ideal | utilization |\n"
-        result += "| --- | --- | --- | --- | --- | --- | --- | --- | --- |n"
+    for layout, add_c in itertools.product(("cyclic", "banked"), (True, False)):
+        result += f"Results for a {layout} layout {'with add C' if add_c else ''} \n\n"
+        result += "| benchmark | layout | add C | M | N | K | plots | cycles | ideal | utilization |\n"
+        result += "| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |\n"
```
JosseVanDelm, line 99: I have just rediscovered pandas dataframes and I really regret not using them sooner, together with seaborn it allows you to do some very fast and beautiful (interactive I think even) plotting. Should also make it a lot easier to create tables such as the ones here. Not for this PR

compiler/transforms/clear_memory_space.py
```diff
@@ -8,28 +9,28 @@
 class ClearMemorySpace(ModulePass):
     name = "clear-memory-space"
 
-    def apply(self, ctx: MLContext, module: builtin.ModuleOp) -> None:
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
         # helper function to clear the memory space of a memref
         # also clears the layout information of the memref - not used anymore
         def clear_memory_space(t):
-            if isinstance(t, memref.MemRefType):
+            if isinstance(t, builtin.MemRefType):
```
JosseVanDelm, line 16: I don't understand why this is necessary to change if we haven't updated xdsl

compiler/transforms/clear_memory_space.py
```diff
@@ -46,3 +47,18 @@ def clear_memory_space(t):
...
+                        BlockArgument(
+                            clear_memory_space(old_arg.type),
+                            op_in_module.body.block,
+                            index,
+                        )
+                        for index, old_arg in enumerate(old_args)
+                    ]
+                    for old_arg, new_arg in zip(old_args, new_args):
+                        old_arg.replace_by(new_arg)
+                    op_in_module.body.block._args = tuple(new_args)
```
JosseVanDelm, line 64: This is to change the function body? or ?

JosseVanDelm (approved):
I have a few questions, but looks super nice overall.
Sorry about the mnk mess

benchmarks/dense_matmul/genbenchmark.py
jorendumoulin: Yes, that is indeed probably a much better solution, they ven seem to have `.to_markdown` functions for nice printing

jorendumoulin: seaborn is probably not allowed on github pages, idk

compiler/transforms/clear_memory_space.py
jorendumoulin: It's not necessary, but memreftype is defined in `builtin`, not in `memref`. 
This is just to not have this message screaming at me:
image

I know it's not really related to this PR, but I tend to just resolve these kinds of things when I notice them and then sneak them in

benchmarks/dense_matmul/main.c
```diff
@@ -15,8 +15,8 @@
 #include <snrt.h>
 
 // Kernel provided via external definition
-void _mlir_ciface_streamer_matmul(TwoDMemrefI8_t *a, TwoDMemrefI8_t *b,
-                                  TwoDMemrefI32_t *c);
+void _mlir_ciface_streamer_matmul(TwoDMemrefI32_t *d, TwoDMemrefI8_t *a,
+                                  TwoDMemrefI8_t *b, TwoDMemrefI32_t *c);
```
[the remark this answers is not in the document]
jorendumoulin: c is always passed, just not always used.

compiler/transforms/clear_memory_space.py
jorendumoulin: to change the types of the block arguments of the funciton body (remove memory space).
Because this is the first time there is an unused function argument, this block argument was not converted to not have memory spaces defined, so this had to be resolved

compiler/transforms/clear_memory_space.py
JosseVanDelm: Oh okay yeah sure no problem at all :)

PR #288, opened 2024-10-15
initialize classes for the scheduling mechanism

jorendumoulin:
This PR provides a significant clean up of the `schedule-memref-linalg` function.

This PR instantiates classes to describe access patterns, templates and schedules, along with a nice test suite.

The algorithm itself is completely unchanged, but already looks a lot simpler.


compiler/ir/stream/access_pattern.py
```diff
@@ -0,0 +1,242 @@
...
+    """
+
+    def __init__(self, bounds: Sequence[int], pattern: AffineMap):
+        super().__init__(bounds, pattern)
+
+    def tile_dim(self, dim: int, template_bound: int) -> Self:
+        raise RuntimeError("A template should not be tiled")
+
+    def rotate(self, dim: int) -> Self:
+        raise RuntimeError("A template should not be rotated")
```
JosseVanDelm, line 165: Instead of raising a RuntimeError, can we not just not implement these methods for this class, then Pyright will complain  whenever you want to call this?

compiler/ir/stream/access_pattern.py
```diff
@@ -0,0 +1,242 @@
...
+    def __init__(self, patterns: Iterable[P]):
+        self._patterns = tuple(patterns)
+
+    @overload
+    def __getitem__(self, index: int) -> P:
+        ...
+
+    @overload
+    def __getitem__(self, index: slice) -> tuple[P]:
+        ...
```
JosseVanDelm, line 197: oof, is this how you're supposed to tell pyright? Cool! but oof!?

compiler/ir/stream/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .access_pattern import *
+from .scheduler import *
```
JosseVanDelm, line 2: For my information: Is this an import helper?

JosseVanDelm (commented):
This looks much much cleaner than before!
However, I'm not fully on board with the schedule implementation, not entirely sure why though.

compiler/ir/stream/access_pattern.py
jorendumoulin: yes: https://docs.python.org/3/library/typing.html#typing.overload
but i find this quite nice? how else would you specify specific different return types for different input types

compiler/ir/stream/access_pattern.py
jorendumoulin: yes that's probably better, I will modify it to this behaviour!

compiler/ir/stream/access_pattern.py
```diff
@@ -0,0 +1,242 @@
...
+
+        if pattern.num_symbols > 0:
+            raise ValueError("Symbols in the pattern are not supported")
+
+        # Canonicalize the pattern
+        new_pattern = canonicalize_map(pattern)
+
+        # Assign attributes using object.__setattr__ due to frozen=True
+        object.__setattr__(self, "bounds", bounds)
+        object.__setattr__(self, "pattern", new_pattern)
```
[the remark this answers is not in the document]
jorendumoulin: attrs seem nice, but is it worth adding a whole dependency just for this?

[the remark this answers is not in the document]
JosseVanDelm: No no no !

compiler/ir/stream/__init__.py
jorendumoulin: this way you can do 
```
from compiler.ir.stream import Template
```
instead of having to do
```
from compiler.ir.stream.access_pattern import Template
```

JosseVanDelm (approved):
yay :heart:

PR #289, opened 2024-10-16
add scheduler tests

jorendumoulin:
This adds some tests to the streamer scheduler for the currently supported cases, and includes little bugfixes for some cases I thought would be currently supported but apparently were not.


tests/ir/stream/test_access_pattern.py
jorendumoulin, line 1: the diffs in this file just move some tests to a more logical ordering

JosseVanDelm (approved):

PR #294, opened 2024-11-28
fix alloc kernel

### LINKED ISSUE #260: alloc kernel is failing
somehow it is not included in the ci, and it is also failing locally for me

jorendumoulin:
There was a check for alignment to 256 bytes, but 64 bytes is sufficient.

Also add the kernel to ci

Resolves #260


jorendumoulin:
@JosseVanDelm

JosseVanDelm (approved):

PR #295, opened 2024-12-03
new bufferization strategy

jorendumoulin:
This PR changes the order of bufferization:

Previously:

1. bufferize custom ops
2. --one-shot-bufferize

Now:

1. --one-shot-bufferize ( + allow_custom_ops)
2. bufferize custom ops

This works much better with more complex structures such as tiled programs, as the --one-shot bufferize can use tensor semantics to better determine when and when not to create new buffers.

The `snax-bufferize` and `stream-bufferize` passes are also merged to `snax-bufferize`


JosseVanDelm (approved):
Nice!

PR #296, opened 2024-12-03
update matmul kernel

jorendumoulin:
note: stacked PR, awaits #295 

This changes the matmul test to be generated by a python file with data included, rather than having the logic of the test distributed across gendata, the mlir file and the main.c file.

It also starts from the tensor level rather than the memref level, and includes a bufferization pass.

PR #297, opened 2024-12-03
bump xdsl, mlir and snax

jorendumoulin:
main changes in xdsl:
- renaming all ops to end with `Op` -> fixed by renaming
- no rewritepatterns allowed on `builtin.ModuleOp` -> fixed by moving the functionality of these passes to the modulepass instead of using a rewritepattern

main changes in mlir:
- arith.overflow attributes fucking up a lot of filechecks -> fixed by adding these attributes (if there were a few) or converting the filecheck to not print in generic format

main changes in snax:
- broadcasting and rewrite address mapping functionalities in the gemmx, most of them are just hardcoded now

The only thing I haven't gotten working is how to call the new transform interpreter correctly.
However, as this was going to change anyhow, I just disabled it for now and should be back up and running with #296


JosseVanDelm (approved):
Not all heroes wear capes.
Great work! Thanks!

PR #301, opened 2024-12-16
add pixi python initialization

jorendumoulin:
Initializes the pixi properties for the python packages.

This currently restricts the python version to `3.10.*`, something that can be updated further down the line


JosseVanDelm (approved):
:fire:

PR #302, opened 2024-12-16
pixi: add mlir, clang, lld

jorendumoulin:
this adds dependencies to the (approximately) same commit as xdsl is based on.
Locally with very little changes (paths to binaries and so on), and the /opt folder from the docker container copied locally,
this compiled and simulated a kernel successfully!


JosseVanDelm (approved):
Yes!

PR #304, opened 2024-12-17
update ci to pixi

jorendumoulin:
important to use ubuntu 24 for glibc 2.38!


pyproject.toml
```diff
@@ -17,6 +17,7 @@ dev = [
     "lit",
     "numpy<2.0.0",
     "pytest",
+    "pyright",
```
JosseVanDelm, line 20: Can we do this in a PR that also adds pyright to pre-commit?

JosseVanDelm (approved):
Good! Can we keep pyright for your personal environment for a short while?

Also:
> important to use ubuntu 24 for glibc 2.38!
Why? Is this because this is the version we have used to build these packages on our systems with rattler build? Seems like we should be careful isolating these?

jorendumoulin:
https://conda-forge.org/news/2024/11/22/new-images/
this is how to do the glibc versioning better if we run into any issues

PR #309, opened 2024-12-18
fix pyright for dialects/stream and util/dispatching_rules

jorendumoulin:
this fixes pyright for the following files:

- dialects/stream
- util/dispatching-rules


JosseVanDelm (approved):
Nice!

PR #310, opened 2024-12-18
Add alloc snakemake kernel

JosseVanDelm:
This PR adds the basic infrastructure to run a snakemake experiment for the alloc kernel, I will update the other kernels with different PRs :)


kernels/alloc/Snakefile
```diff
@@ -0,0 +1,47 @@
...
+
+
+rule all:
+    input:
+        "func.x",
+    shell:
+        "{config[vltsim]} {input[0]}"
+
+
+use rule * from default_rules exclude compile_simple_main as default_*
```
jorendumoulin, line 37: what is this?

jorendumoulin (approved):
😍

kernels/alloc/Snakefile
JosseVanDelm: This is to include/import rules from a different snakemake file declared as a module.
You use the module declaration to choose the file and to apply the configuration, and then those rules act as if they have been included in the file, but with the `default_` prefix.
In this case, a lot of rules are shared for many jobs, so instead of defining them everywhere multiple times, I import them from `default_rules.smk`

jorendumoulin: but why exclude compile_simple_main?

PR #311, opened 2024-12-18
Snakemake: Add support for transform-copy

JosseVanDelm:
Runs transform-copy with snakemake instead of makefiles

Awaits #310


.github/workflows/build-run-kernel-snake.yml
```diff
@@ -16,8 +16,8 @@ jobs:
           cache: true
           cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
       - name: Build and run kernels
-        run: pixi run snakemake -k -j `nproc`
+        run: pixi run snakemake -k -j `nproc` all
```
jorendumoulin, line 19: could we maybe make a pixi task for this such that we can do `pixi make`? would be convenient
not needed here though, just a thought

jorendumoulin (approved):

.github/workflows/build-run-kernel-snake.yml
JosseVanDelm: what is a pixi-task? :D

PR #313, opened 2024-12-19
Snakemake: Add gemm kernel

JosseVanDelm:
Adds gemm kernel and gemmx config


kernels/gemm/Snakefile
```diff
@@ -0,0 +1,91 @@
...
+
+
+module default_rules:
+    snakefile:
+        "../../util/snake/default_rules.smk"
+    config:
+        config
+
+
+use rule * from default_rules exclude mlir_preprocessing as default_*
```
jorendumoulin, line 41: again here: where is mlir_preprocessing?

jorendumoulin (approved):

kernels/gemm/Snakefile
JosseVanDelm: Good catch!

PR #316, opened 2024-12-19
Fix gemmini kernel, add snakemake, remove Makefile

JosseVanDelm:
There was an error in changing the upstream MLIR kernel that was not caught by CI, here's an updated version that uses Snakemake


kernels/gemmini/Snakefile
```diff
@@ -0,0 +1,86 @@
...
+    snakefile:
+        "../../util/snake/default_rules.smk"
+    config:
+        config
+
+
+use rule preprocess_mlir from default_rules as default_preprocess_mlir
+
+
+use rule translate_mlir from default_rules as default_translate_mlir
```
jorendumoulin, line 40: why the renaming?

jorendumoulin (approved):

kernels/gemmini/Snakefile
JosseVanDelm: This is to indicate that we import them from the defaults.
```
use rule preprocess_mlir, translate_mlir from default_rules as default_*
```
Also works, but snakefmt wouldn't let me, maybe the newer versions fix this?

PR #322, opened 2025-01-02
add affine transform util

jorendumoulin:
An AffineTransform helper, similar to the implementation in MLIR/xDSL.
However, the representation is handled by a matrix A and B instead of the AffineExpr tree, which should make things much easier. I plan to use this class inside the `compiler/ir/stream/access_pattern.py` utility classes.
This also gets rid of the canonicalization problem for affine transforms.

### DIFF
--- a/compiler/ir/autoflow/__init__.py
+++ b/compiler/ir/autoflow/__init__.py
@@ -1,0 +1,1 @@
+from .affine_transform import *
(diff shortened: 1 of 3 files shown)


JosseVanDelm:
> An AffineTransform helper, similar to the implementation in MLIR.

Can you link it please?

jorendumoulin:
> > An AffineTransform helper, similar to the implementation in MLIR.
> 
> Can you link it please?

there isn't a lot of useful information here but here you go: https://github.com/llvm/llvm-project/blob/main/mlir/lib/IR/AffineMap.cpp

JosseVanDelm (approved):
sweet!

PR #324, opened 2025-01-02
affine transform util: add interoperability with AffineMaps

jorendumoulin:
to keep interoperability with xdsl afinemaps, and to enable a more gradual migration to affinetransforms, include the constructors `from_affine_map` and `to_affine_map`

### DIFF
--- a/compiler/ir/autoflow/affine_transform.py
+++ b/compiler/ir/autoflow/affine_transform.py
@@ -3,6 +3,7 @@ from dataclasses import dataclass
 import numpy as np
 import numpy.typing as npt
 from typing_extensions import Self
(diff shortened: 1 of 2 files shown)


JosseVanDelm (changes requested):
Tests are looking great, but could you add a bit more descriptive comments?

E.g. right now it is quite difficult to read what is going on.

For example, what does generate_one_list do?
Something like:
```
construct a matrix representation of the affine expression by evaluating the dimensions one by one...
```
Also, can you add a check such that the input affine expr does not contain any of the invariants (floordiv/ceildiv)?

JosseVanDelm (approved):
Thank you @jorendumoulin !!!

PR #326, opened 2025-01-03
pyright: fix custom dialects snax, snax_stream, tsl

jorendumoulin:


JosseVanDelm (approved):
Great that we are stepping away from i32 on snax.alloc!
Good job!

PR #329, opened 2025-01-03
remove linalg-to-library-call pass

jorendumoulin:
It isn't used anymore, accfg lowering should be used from now on


JosseVanDelm (approved):
:fire: Thank you :pray: !

PR #332, opened 2025-01-06
split convert_stream_to_snax_stream in scheduler and layout resolution

jorendumoulin:
stacked on #331 

With the addition of the schedule op, this PR splits the `convert_stream_to_snax_stream` (which converts operation -> access pattern) into two separate passes (scheduler: operation -> schedule), (layout resolution: schedule -> access pattern).

The schedule representation is nice to have to run other passes on before doing layout resolution, such as selecting a layout.

### DIFF
--- a/compiler/transforms/convert_stream_to_snax_stream.py
+++ b/compiler/transforms/convert_stream_to_snax_stream.py
@@ -13,55 +13,23 @@ from xdsl.pattern_rewriter import (
     op_type_rewrite_pattern,
 )
 
-from compiler.accelerators import find_accelerator_op
-from compiler.accelerators.registry import AcceleratorRegistry
-from compiler.accelerators.snax import SNAXStreamer
 from compiler.dialects import snax_stream, stream
 from compiler.ir.stream import Schedule, SchedulePattern, scheduler
-from compiler.ir.stream.access_pattern import Template
-
-
-def get_accelerator_info(op: stream.StreamingRegionOpBase) -> Template:
-    assert op.accelerator is not None
-
-    # Go and fetch the accelerator op
-    accelerator_str = op.accelerator.data
-    acc_op = find_accelerator_op(op, accelerator_str)
-
-    if not acc_op:
-        raise RuntimeError("AcceleratorOp not found!")
-
-    # get template and template_bounds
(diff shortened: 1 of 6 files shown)


compiler/transforms/convert_stream_to_snax_stream.py
```diff
@@ -17,12 +17,31 @@
 from compiler.accelerators.registry import AcceleratorRegistry
 from compiler.accelerators.snax import SNAXStreamer
 from compiler.dialects import snax_stream, stream
-from compiler.dialects.snax import StreamerConfigurationAttr
 from compiler.ir.stream import Schedule, SchedulePattern, scheduler
+from compiler.ir.stream.access_pattern import Template
+
+
+def get_accelerator_info(op: stream.StreamingRegionOpBase) -> Template:
```
JosseVanDelm, line 24:
```suggestion
def get_accelerator_template(op: stream.StreamingRegionOpBase) -> Template:
```

JosseVanDelm: Can you make this a method of the (base class of the) op itself maybe?

compiler/transforms/convert_stream_to_snax_stream.py
```diff
@@ -77,6 +76,31 @@ def match_and_rewrite(
...
+        )
+
+        rewriter.replace_matched_op(schedule_op)
+
+
+@dataclass
+class LayoutResolution(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: stream.ScheduleOp, rewriter: PatternRewriter):
+        template = get_accelerator_info(op)
```
JosseVanDelm, line 97: If this were a method:
```suggestion
        template = op.get_accelerator_template()
```

JosseVanDelm (approved):
I have some small organizational proposed changes.
accfg registry is definitly a bit messy in this regard, which is a shame, but also no deal-breaker right now.
Can you also maybe update the comments a bit?
I guess more changes are also coming in future PRs, so please do when you see fit!

compiler/transforms/convert_stream_to_snax_stream.py
jorendumoulin: unfortunately this would be a circular import

PR #334, opened 2025-01-09
transforms: (set_memory_layout) only apply on streaming_regions

jorendumoulin:
I want to operate this pass only on schedules (after the autoflow scheduler), where the layout is determined automatically according to the access pattern to memory.

This is a first step, to make sure everything keeps working when setting the memory layout at this stage (instead of at linalg.generic level)


compiler/transforms/set_memory_layout.py
```diff
@@ -127,41 +120,29 @@ class AddMemoryLayout(RewritePattern):
...
-            if isinstance(op, linalg.GenericOp):
-                if not isinstance(op.body.block.first_op, QMacOp):
-                    return
-            elif isinstance(op, stream.StreamingRegionOp):
-                assert isinstance(
-                    generic_op := op.body.block.first_op, stream.GenericOp
-                )
-                if not isinstance(generic_op.body.block.first_op, QMacOp):
-                    return
+            assert isinstance(generic_op := op.body.block.first_op, stream.GenericOp)
```
JosseVanDelm, line 138: nit: Feels a bit weird to assign a variable inside an assertion?
```suggestion
            generic_op = op.body.block.first_op
            assert isinstance(generic_op, stream.GenericOp)
```

JosseVanDelm (approved):
That makes perfect sense! Thanks!

PR #339, opened 2025-01-20
get rid of tollvm12 script

jorendumoulin:
we don't use llvm12 for quite some time now


JosseVanDelm:
I'm actually not fully sure we have double precision float support now?

JosseVanDelm (approved):
Okay, I guess?

PR #340, opened 2025-01-20
integrate preprocessing in snax-opt

jorendumoulin:
stacked on #341 

This incorporates the preprocessing passes in snax-opt, such that we don't need to mess around with separate mlir-opt preprocessing. And `preproc1`, `preproc2` and `preprocfinal` files.


JosseVanDelm (approved):
Very nice!

PR #346, opened 2025-01-22
split up dart transformations

jorendumoulin:
This splits the three patterns into separat transformations. This will allow me to:

- Test the intermediate stages more properly
- Put other transforms in between that act on the intermediate representations

### DIFF
--- a/compiler/tools/snax_opt_main.py
+++ b/compiler/tools/snax_opt_main.py
@@ -28,6 +28,8 @@ from compiler.transforms.convert_linalg_to_kernel import ConvertLinalgToKernel
 from compiler.transforms.convert_tosa_to_kernel import ConvertTosaToKernelPass
 from compiler.transforms.dart.convert_linalg_to_dart import ConvertLinalgToDart
 from compiler.transforms.dart.dart_fuse_operations import DartFuseOperationsPass
+from compiler.transforms.dart.dart_layout_resolution import DartLayoutResolutionPass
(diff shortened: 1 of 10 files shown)


compiler/transforms/convert_dart_to_snax_stream.py
JosseVanDelm, line 1: Should this also be part of the dart folder?

compiler/tools/snax_opt_main.py
```diff
@@ -28,6 +28,8 @@
 from compiler.transforms.convert_tosa_to_kernel import ConvertTosaToKernelPass
 from compiler.transforms.dart.convert_linalg_to_dart import ConvertLinalgToDart
 from compiler.transforms.dart.dart_fuse_operations import DartFuseOperationsPass
+from compiler.transforms.dart.dart_layout_resolution import DartLayoutResolutionPass
+from compiler.transforms.dart.dart_scheduler import DartSchedulerPass
```
JosseVanDelm, line 32: I think we need an __init__.py file in the dart folder for this to work in external (pip) installations

JosseVanDelm (approved):
Looks great! Can you add an init?

compiler/transforms/convert_dart_to_snax_stream.py
jorendumoulin: I don't think so, it's more related to streamers

compiler/tools/snax_opt_main.py
jorendumoulin: there already is

PR #347, opened 2025-01-23
Run simulations in parallel

JosseVanDelm:
This adds --prefix-trace support and several utilities to handle the quite difficult naming scheme provided by the simulation tracer.


util/snake/paths.py
```diff
@@ -6,3 +9,29 @@ def get_default_paths():
...
         "snax-opt": "snax-opt",
     }
+
+
+def get_traces(
+    prefixes: Sequence[str],
+    num_chips: int = 1,
+    num_harts: int = 2,
+    extension: str = "dasm",
+) -> Generator[str, None, None]:
```
jorendumoulin, line 19: is this type correct?

util/snake/paths.py
```diff
@@ -6,3 +9,29 @@ def get_default_paths():
         "mlir-translate": "mlir-translate",
         "snax-opt": "snax-opt",
     }
+
+
+def get_traces(
```
jorendumoulin, line 14: why is this defined in `/paths`?

util/snake/default_rules.smk
```diff
@@ -54,3 +57,20 @@ rule clean:
...
+rule simulate:
+    input:
+        "{file}.x",
+    output:
+        temp(
+            multiext(
+                *get_trace_ext(
+                    "{file}", config["num_chips"], config["num_harts"], "dasm"
+                )
+            )
```
jorendumoulin, line 71: > multiext provides a simplified variant of expand that allows us to define a set of output or input files that just differ by their extension:

why would traces just differ by their extension? I don't think they do?

util/snake/paths.py
JosseVanDelm: Well, because it's just a small utility that generates filenames, that's part of a path, so ...
I can definitely put it somewhere else if you like :)

util/snake/paths.py
JosseVanDelm: It looks very weird indeed! But I hope I annotated it correctly :sweat_smile: 
There's no return statement in this function, there's nothing that reads from the yield, so I guess so:
https://peps.python.org/pep-0484/#annotating-generator-functions-and-coroutines

util/snake/default_rules.smk
JosseVanDelm: solved in 0b5c72d735bdadb093fe89dc28f83e68e4121ac5

util/snake/paths.py
JosseVanDelm: Removed now!

util/snake/paths.py
JosseVanDelm: Not necessary anymore, using `expand()` now!

util/snake/snax.smk
```diff
@@ -0,0 +1,16 @@
...
+        temp(
+            expand(
+                "{file}_trace_chip_{num_chips:02d}_hart_{num_harts:05d}.dasm",
+                file=["{file}"],
+                num_chips=range(config["num_chips"]),
+                num_harts=range(config["num_harts"]),
+            ),
+        ),
+    log:
+        "{file}.vltlog",
```
jorendumoulin, line 14: won't this be put in the /logs folder or something?

jorendumoulin:
I'm sorry to be so annoying for this, but I don't like that we need to specify all of the dasm targets in every test separately.
Could we maybe already complete the chain like this?

```
   rule: simulate    rule: trace    rule: aggregate_json 
                                                         
                                                         
     ┌───► .._0.dasm  ────►  .._0.json ───────►          
     │                                                   
     │ ┌─► .._1.dasm  ────►  .._1.json ───────► file.json
file.x │                                                 
      │    .._2.dasm ─────►  .._2.json  ──────►          
      └───►                                              
```

Such that all of the rules in the _downstream_ Snakefiles simplify to:
```
rule all:
    input: file.json
```

util/snake/snax.smk
JosseVanDelm: @jorendumoulin, no that was (and still actually is) the original default. 
So if you don't specify `--prefix-trace` it will generate the logs folder, if you specify anything else, it should not do that (unless you specify something like `"my_first_little_logs_folder/my_prefix"`)

JosseVanDelm:
@jorendumoulin  that is a great idea! I'll work on it!

kernels/gemm/Snakefile
```diff
@@ -39,6 +39,16 @@ module default_rules:
...
+module snax_rules:
+    snakefile:
+        "../../util/snake/snax.smk"
+    config:
+        config
+
+
+use rule * from snax_rules as snax_*
+
+
```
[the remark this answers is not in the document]
JosseVanDelm: That's a very good suggestion! It hadn't crossed my mind yet :) . I'll try to see if that works!

util/snake/paths.py
```diff
@@ -6,3 +9,13 @@ def get_default_paths() -> dict[str, str]:
         "mlir-translate": "mlir-translate",
         "snax-opt": "snax-opt",
     }
+
+
+def get_default_snax_paths() -> dict[str, str]:
+    # use CONDA_PREFIX to access pixi env
+    gen_trace_path = f"{os.environ['CONDA_PREFIX']}/snax-utils/gen_trace.py"
+    return {
+        "python": "python",
```
jorendumoulin, line 18: is this snax-specific?

util/snake/paths.py
```diff
@@ -6,3 +9,13 @@ def get_default_paths() -> dict[str, str]:
         "mlir-translate": "mlir-translate",
         "snax-opt": "snax-opt",
```
jorendumoulin, line 10: is this not snax-specific?

util/snake/default_rules.smk
```diff
@@ -48,6 +48,20 @@ rule compile_llvm_module:
...
+    input:
+        "{file}.dasm",
+    output:
+        temp("{file}.json"),
+        temp("{file}.txt"),
+    shell:
+        "{config[cc]} {config[clangflags]} -x ir -c {input} -o {output}"
+
+
 rule clean:
```
jorendumoulin: delete

kernels/gemm/Snakefile
```diff
@@ -51,11 +51,13 @@ rule compile_main:
...
+files = ["gemm"]
+
+
+# Rules
 rule all:
     input:
-        "gemm.x",
-    shell:
-        "{config[vltsim]} {input[0]}"
+        expand("{file}_aggregated.json", file=files),
```
jorendumoulin:
```suggestion
        expand("{file}_traces.json", file=files),
```
?

jorendumoulin: to simplify (and others)

jorendumoulin (approved):
yay 😍

PR #349, opened 2025-01-24
add access pattern inner_dims method

jorendumoulin:
This is meant to replace `disable_dims`, as it is much more intuitive and easier to use. The algorithms are still processing this change, so `disable_dims` is just deprecated for now.


JosseVanDelm (approved):
Okay! Let's deprecate disable_dims asap!

PR #350, opened 2025-01-27
add backtracking scheduler

jorendumoulin:
stacked on #351 -> #349 

This PR implements a nicer scheduler using a backtracking algorithm to generate valid schedules on the operation.

### DIFF
--- a/compiler/ir/dart/scheduler.py
+++ b/compiler/ir/dart/scheduler.py
@@ -1,43 +1,90 @@
+from collections.abc import Generator
+
 from compiler.ir.dart.access_pattern import Schedule, Template
 
 
-def scheduler(template: Template, schedule: Schedule) -> Schedule:
-    for i in range(template.num_dims):
-        # i = 0: look at the last dimension
-        # i = 1: look at the second to last dimension
-        template_dim = template.num_dims - i - 1
-        schedule_dim = schedule.num_dims - i - 1
-        match = False
+def scheduler_backtrack(
+    template: Template,
+    schedule: Schedule,
+    inner_dims: int = 1,
+) -> Generator[Schedule]:
+    """
+    Backtracking method to find all possible mappings of the schedule on the template
 
-        # maximum number of rotations
-        for _ in range(schedule_dim + 1):
-            # check if there is a match
-            template_check = template.disable_dims(template_dim)
-            schedule_check = schedule.disable_dims(schedule_dim)
+    `template` (Template): the accelerator template
+    `schedule` (Schedule): the partially scheduled operation
+    `inner_dims` (int): current number of innermost dimensions being handled
+    `pure_output_stationary` (bool):
+    """
 
-            if template_check.matches(schedule_check):
-                match = True
-                break
+    """
+    Explanation of dimensions:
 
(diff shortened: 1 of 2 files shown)


compiler/ir/dart/scheduler.py
jorendumoulin, line 1: do not try to understand the diff, this file is completely new and rewritten

compiler/ir/dart/scheduler.py
```diff
@@ -1,43 +1,90 @@
...
+
+    """
+    Explanation of dimensions:
+
+        In case we are handling 6 dimensions and `dim` = 3:
+
+        There are 3 innermost dimensions that are being checked with the template.
+        The other outermost dimensions are not considered.
+
+        The current dimension under foces (d3) is the most outermost dim of the innermost dims.
```
JosseVanDelm, line 28: I would write as
```suggestion
        The currently focused dimension (d3) is the most outermost dim of the innermost dims.
```
or 
```suggestion
        The current dimension in question (d3) is the most outermost dim of the innermost dims.
```
or
```suggestion
        The current dimension under consideration (d3) is the most outermost dim of the innermost dims.
```

tests/ir/dart/test_access_pattern.py
JosseVanDelm, line 1: This seems to not be properly stacked as you say though @jorendumoulin ?

compiler/ir/dart/scheduler.py
```diff
@@ -1,43 +1,90 @@
...
+                )
+
+        # continue with candidate schedule, with an extra inner dim:
+        yield from scheduler_backtrack(template, candidate_schedule, inner_dims + 1)
+
+
+def scheduler(template: Template, schedule: Schedule) -> Schedule:
+    # for now just return the first result of the backtracking
+    result = next(scheduler_backtrack(template, schedule))
+    return result
```
JosseVanDelm, line 90: Can we test somewhere what happens if there are multiple valid schedules?

JosseVanDelm (approved):
Looks good modulo some comments

compiler/ir/dart/scheduler.py
```diff
@@ -1,43 +1,90 @@
...
-            raise RuntimeError("failed to match template and schedule")
-
-        # now, check bounds and design potential transformation map
-        if not (template_bound := template[0].bounds[template_dim]):
-            # nothing to worry about, continue to next dim
+def scheduler_backtrack(
+    template: Template,
+    schedule: Schedule,
+    inner_dims: int = 1,
+) -> Generator[Schedule]:
```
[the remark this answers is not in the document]
jorendumoulin: Indeed, but after some reading I think more accurately this is an `Iterator`, and not a Generator (which should have the methods __send__ and return, that do not apply here.

tests/ir/dart/test_access_pattern.py
jorendumoulin: i don't understand this comment?
However, these changes are actually part of #351, i messed up my rebasing a bit it seems

jorendumoulin: ah no the base was set incorrectly

PR #351, opened 2025-01-27
template matching: only look at the inner dims of the schedule

jorendumoulin:
stacked on #349 

with this PR, when doing a template check with a template and a schedule, previously they are forced to the same number of dims. This PR checks if the schedule has larger dimensionality than the template, and reduces it to the number of dims of the template in this case.

### DIFF
--- a/compiler/ir/dart/access_pattern.py
+++ b/compiler/ir/dart/access_pattern.py
@@ -198,7 +198,7 @@ class TemplatePattern(AccessPattern):
         """
         if sp.num_dims > self.num_dims:
             sp = sp.inner_dims(self.num_dims)
-        if sp.num_dims != self.num_dims:
+        elif sp.num_dims < self.num_dims:
             return False
         if sp.pattern != self.pattern:
             return False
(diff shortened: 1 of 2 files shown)


tests/ir/dart/test_access_pattern.py
```diff
@@ -226,20 +226,34 @@ def test_template_pattern_matches():
...
+    # test non matching pattern
     sp_non_matching_pattern = SchedulePattern(
         bounds,
         AffineMap(
             num_dims=2, num_symbols=0, results=(AffineDimExpr(1), AffineDimExpr(0))
         ),
     )
+    assert tp.matches(sp_non_matching_pattern) is False
+
+    # check pattern with wrong bounds (should be irellevant for template check)
```
JosseVanDelm, line 243:
```suggestion
    # check pattern with wrong bounds (should be irrelevant for template check)
```

compiler/ir/dart/access_pattern.py
```diff
@@ -196,6 +196,8 @@ def matches(self, sp: SchedulePattern):
         Check if a given schedule pattern matches this
         template pattern.
         """
+        if sp.num_dims > self.num_dims:
+            sp = sp.inner_dims(self.num_dims)
         if sp.num_dims != self.num_dims:
```
JosseVanDelm, line 201: maybe make this more explicit?
```suggestion
        elif sp.num_dims < self.num_dims:
```

JosseVanDelm (approved):
Looks good!

PR #354, opened 2025-01-28
add optional output stationary check

jorendumoulin:
stacked on #350 

This one adds the capability to the scheduler to add extra checks/ constraints to filter out unwanted schedules, and adds a check to only accept pure output stationary (= never send partial results back to memory) schedules.

This is a useful  check for for example accelerators which do not have the capability to preload partial results. In our case the codegen further down the pipeline is not yet capable of non-output stationary schedules.

Currently, this constraint is embedded in accelerator templates in a hacky way. This way of handling things makes it more explicit and controllable.

### DIFF
--- a/compiler/ir/dart/scheduler.py
+++ b/compiler/ir/dart/scheduler.py
@@ -98,7 +98,7 @@ def is_pure_output_stationary(template: Template, schedule: Schedule):
     """
     Checks whether a schedule, outside of the template, is fully output
     stationary. This is determined by making sure all parallel dimensions
-    preced the reduction dimensions in the output operand (last operand).
+    precede the reduction dimensions in the output operand (last operand).
     """
     # fetch the pattern of the last operand
     output_schedule = schedule[-1].pattern.A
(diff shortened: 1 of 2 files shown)


compiler/ir/dart/scheduler.py
```diff
@@ -81,10 +89,45 @@ def scheduler_backtrack(
...
+        yield from scheduler_backtrack(
+            template, candidate_schedule, inner_dims + 1, extra_checks
+        )
+
+
+def is_pure_output_stationary(template: Template, schedule: Schedule):
+    """
+    Checks whether a schedule, outside of the template, is fully output
+    stationary. This is determined by making sure all parallel dimensions
+    preced the reduction dimensions in the output operand (last operand).
```
JosseVanDelm, line 101:
```suggestion
    precede the reduction dimensions in the output operand (last operand).
```

compiler/ir/dart/scheduler.py
```diff
@@ -7,6 +9,7 @@ def scheduler_backtrack(
     template: Template,
     schedule: Schedule,
     inner_dims: int = 1,
+    extra_checks: Sequence[Callable[[Template, Schedule], bool]] = [],
```
JosseVanDelm, line 12: :face_holding_back_tears:

JosseVanDelm (approved):
This is beautiful

PR #355, opened 2025-01-28
apply set-memory-layout on dart schedules

jorendumoulin:
a schedule contains all the right information to determine good memory layouts. (this is still manually hardcoded, an automation for this is incoming...)
This PR simply changes the op on which the operation acts as well as the ordering of passes to make this work.

it was important to have `dart-layout-resolution` after the insert-sync-barrier and realize memref casts passes


jorendumoulin:
encountering some bugs here, being solved in #355

JosseVanDelm (approved):
Nice!

PR #357, opened 2025-01-30
Add lazy dialect loading

JosseVanDelm:
This is a first refactoring towards making something like `snax-runner`,
I'm refactoring the way our dialects get loaded to be more in line with how xdsl is doing it.
This gives us the benefit of quickly loading dialects in many different python files.
Furthermore it seems to enable lazy loading, which means that the context does not import the dialect definition unless it is really necessary, which is pretty nice!


jorendumoulin:
very nice, although this is not really lazy loading anything?
I slightly prefer how it is done in https://github.com/xdslproject/inconspiquous and https://github.com/xdslproject/xdsl-aie/, where the the `get_all_dialects` is overriden, instead of register_all_dialects. This way we can explicitly note down which dialects to include, rahter than to include all of them, and override some of them

What do you think about that?

compiler/dialects/test/__init__.py
jorendumoulin, line 1: I would just put this one along in the dialects/__init__?

jorendumoulin (approved):

PR #359, opened 2025-01-31
simplify accelerator templates

jorendumoulin:


tests/filecheck/transforms/convert-dart-to-snax-stream.mlir
jorendumoulin, line 1: this schedule changes from (M -> N -> K) to (N -> M -> K), both are valid output stationary schedules, but only the first one was possible before this change. Because we just take the first schedule right now in dart scheduler, the one that is selected in the end is a bit random.

JosseVanDelm (approved):
Okay looks cool!

PR #360, opened 2025-01-31
add simplify method to tiled strided layouts

jorendumoulin:


JosseVanDelm (approved):
Maybe call it canonicalize?

PR #361, opened 2025-02-01
use auto-generated cyclic layouts

jorendumoulin:
stacked on #360


compiler/transforms/set_memory_layout.py
jorendumoulin, line 1: again, this diff is not useful, the entire file has been replaced by a new implementation

JosseVanDelm (approved):
Nice!

PR #363, opened 2025-02-04
add numpy dependency explicitly

jorendumoulin:
We use it for quite some things now


JosseVanDelm (approved):
Sounds like a good idea! Can you fix the lock-file ?

PR #369, opened 2025-02-20
MLContext -> Context

JosseVanDelm:


snaxc/dialects/dart.py
```diff
@@ -1,6 +1,7 @@
 from collections.abc import Iterable, Mapping, Sequence
-from typing import Generic, TypeVar
+from typing import Generic
 
+from typing_extensions import TypeVar
```
JosseVanDelm, line 4: This took me a while to figure out.
The default one in typing is only available from python 3.13

jorendumoulin (approved):

PR #370, opened 2025-02-25
Add acc context mechanism to register accelerators

JosseVanDelm:
This adds an AccContext class which overrides the regular context to add a dictionary with accelerator factories.
In this way other tools or subtools from this project can use these passes with their own registered accelerators directly, as available from the context.

Passes can now query the registered accelerators from the context directly, instead of a python import.
Note that we currently do this with assertions `assert isinstance(ctx, AccContext)` to provide backwards compatibility with the upstream xdsl pass infrastructure without making pyright unhappy.

In our current project, the main contribution of this PR, is that we can register extra accelerators (e.g. in Praxis) without having to all add them to the AcceleratorRegistry in snax-mlir, or that we even can read external files (e.g. hjson) to on-the-fly register new accelerator implementations as we invoke the compiler.
This is surely quite useful in a setting where these things change constantly.


snaxc/tools/snax_opt_main.py
```diff
@@ -38,8 +20,9 @@ def __init__(
         self.available_passes = {}
         self.available_targets = {}
 
-        self.ctx = Context()
+        self.ctx: AccContext = AccContext()
         self.register_all_dialects()
+        self.register_all_accelerators()
```
JosseVanDelm, line 25: This would be the main contribution of this PR, the fact that here we can now choose which accelerators are actually registered :)

snaxc/transforms/insert_accfg_op.py
```diff
@@ -25,11 +25,14 @@ class InsertAccOp(ModulePass):
         snax-opt -p insert-accfg-op{accelerater=snax_hwpe_mult}
     """
 
-    def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
+    def apply(
+        self, ctx: Context, op: builtin.ModuleOp
+    ) -> None:  # pyright: ignore[reportIncompatibleMethodOverride]
```
jorendumoulin, line 30: why is this not compatible?

snaxc/transforms/dispatch_kernels.py
```diff
@@ -100,10 +100,13 @@ class DispatchKernels(ModulePass):
 
     def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
         # find all accelerator ops in the IR
+        assert isinstance(ctx, AccContext)
         accelerators: list[type[DispatchTemplate]] = []
         for accelerator_op in op.ops:
             if isinstance(accelerator_op, AcceleratorOp):
-                accelerator_type = AcceleratorRegistry().get_acc_info(accelerator_op)
+                accelerator_type = ctx.get_acc(
+                    str(accelerator_op.properties["name"])[1:]
```
jorendumoulin, line 108: you can get the name easier, but i forgot how :) something like root_reference or something?

jorendumoulin (commented):
Most of this looks very nice, but why does the Context store `type[Accelerator]` instead of just the Accelerator object? Now there are acc() declarations specified all over the place, and the default arguments of the __init__ are always used and not overwritable.

To me it makes more sense to store actual instances of the accelerator, such that this is not necessary? We can then also use different names for different instances of the same accelerator (for example, `snax_gemmx_8x8x8` and `snax_gemmx_64x64`), which are the same accelerator, just with a different parameter.

snaxc/transforms/insert_accfg_op.py
JosseVanDelm: Oops, this was before I figured out the 
```python
assert isinstance(ctx, AccContext)
```

JosseVanDelm:
@jorendumoulin thanks for your suggestions. 
I definitely agree with you, doing this uncovered a lot of inconsistencies from the work that I had done before.
The accfg registry mechanism right now is definitely a bit of mess, and it will probably remain a bit so even after these fixes.
I'll see what I can do :)

snaxc/accelerators/acc_context.py
```diff
@@ -0,0 +1,107 @@
...
+class AccContext(Context):
+    """
+    Context that additionally allows to register and query accelerators
+    """
+
+    _registered_accelerators: dict[str, Callable[[], type[Accelerator]]] = field(
+        default_factory=dict
+    )
+
+    def clone(self) -> "AccContext":
```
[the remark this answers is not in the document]
JosseVanDelm: I definitely had to double check this, but it seems this is the only correct annotation in this case.

This particular method does not work here.
clone does not:
```python
return self
```
So technically, the method does not return Self, but it returns a new instance of AccContext and hence a type annotation that says `Self` is incorrect here.
image
The problem here is that I can not just reference it as `AccContext` because python does not support forward type references:
image

see:
https://realpython.com/python-type-self/
and 
https://stackoverflow.com/questions/55320236/does-python-evaluate-type-hinting-of-a-forward-reference

I can `from __future__ import annotations`, which solves it:
image

But that introduces new possible errors and the realpython reference seems to not like this for various reasons (it feels like a hack and messes up inherited types, because it flattens all annotations in this file as strings), the stackoverflow reference seems to suggest doing this for performance reasons, but to me it seems more like a hack (the import also has to be at the top of the module etc)
What do you think?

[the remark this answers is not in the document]
JosseVanDelm: I did remove a stray typevar in this file after this comment though, so thanks hahah

snaxc/accelerators/acc_context.py
```diff
@@ -0,0 +1,107 @@
...
+    ) -> tuple[AcceleratorOp, type[Accelerator]]:
+        """
+        Perform a symbol table lookup for the accelerator op in the IR
+        and then get the corresponding the Accelerator interface from
+        the accelerator registry.
+        Returns both the looked up accelerator op and the Accelerator interface
+        """
+        acc_op = find_accelerator_op(module, name)
+        if acc_op is None:
+            raise Exception(
```
AntonLydike, line 77: `raise KeyError(` instead?

snaxc/accelerators/util.py
AntonLydike, line 1: Is this file deleted or just "emptied"? diff is hard to read here.

snaxc/transforms/convert_linalg_to_accfg.py
```diff
@@ -34,24 +33,20 @@ class ConvertLinalgToAcceleratorPattern(RewritePattern):
...
     def match_and_rewrite(self, op: linalg.GenericOp, rewriter: PatternRewriter, /):
         if op.library_call is None:
             return
 
         library_call_name = op.library_call.data
-
-        acc_reg = AcceleratorRegistry()
-        acc_names = acc_reg.get_names()
-        if library_call_name not in acc_names:
+        if library_call_name not in self.ctx.registered_accelerator_names:
```
AntonLydike, line 44: Neat!

AntonLydike (approved):
I wonder if this could eventually be solved by the DLTI interface?

But alas, since that's not implemented in xDSL yet, this appears to be a pretty good approximation for now! Cool stuff!

snaxc/accelerators/util.py
JosseVanDelm: appeared to not be fully deleted. Thanks for the suggestion!

snaxc/accelerators/acc_context.py
JosseVanDelm: It is not really a dictionary KeyError that is happening here. This error indicates specifically that no accfg.AcceleratorOp was put in the module...
A KeyError is a bit misleading here in my opinion?

JosseVanDelm:
@jorendumoulin I changed class instantiations to object instantiations.
I also simplified the methods to access the name of the accelerator in question. What do you think?

snaxc/accelerators/acc_context.py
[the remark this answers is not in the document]
jorendumoulin: ok!

jorendumoulin (approved):
Lovely!

PR #374, opened 2025-03-04
add pipeline transform

jorendumoulin:


jorendumoulin:
replaced by #397 - #400 and beyond

PR #380, opened 2025-03-24
singular value template matching

jorendumoulin:
This replaces the naive check if a template matches a schedule with a bit more of a robust one.
While this one is pretty slow, it seems to work very reliably.

If we were to replace this with a better check, at least this one can serve as a benchmark for matching schedules.

This is mainly necessary for convs where the index space is 4D (NCHW) but the accelerator only considers 2 dimensional inputs and outputs


JosseVanDelm (approved):
Cool!

PR #381, opened 2025-03-25
add conv kernel

jorendumoulin:
this adds a kernel to execute convolutions on the gemmx.

with the convspec, different variations of a convolution can be generated, golden results are computed using tensorflow, and the result is given as a `linalg.conv2d_nchw_fchw`.

In ci, the default is a `16x16x16` output with a kernel size of `16x16x3x3`.
I will expand this in future PRs


JosseVanDelm (approved):
Hell yeah!

PR #382, opened 2025-03-25
better knm generation for gemmx

jorendumoulin:
the current method to determine the k,n and m values to program to the accelerator only work for 3 dimensional stride patterns, which is no longer the case for convolutions


JosseVanDelm (approved):
Nice!

PR #384, opened 2025-03-26
remove zero bias hack

jorendumoulin:
this removes the hack to load in zeros via the d32 port to disable the bias addition. the hack is causing some issues.

if we want to compute a matmul without starting from the initial values already present in the output memref, we should provide a more explicit solution for this case

in the case of the current kernels, they are functional as tcdm is initialized to zeros anyway


JosseVanDelm (approved):
Okay!
Can you also remove the extra allocations for zero space in the C runtime for this PR please?

PR #385, opened 2025-03-26
set-memory-layout: only tile if it is nicely divisible

jorendumoulin:
otherwise the layout transformations get a bit messed up.
this is probably something we want to support in the future, but don't right now.


JosseVanDelm (approved):
Nice!

PR #388, opened 2025-03-27
perform correct spatial stride calculation

jorendumoulin:
instead of hardcoding spatial strides, this performs the actual necessary calculations
this is needed for the next platform where serialization of the output streamer is achieved


tests/filecheck/transforms/convert-dart-to-snax-stream.mlir
```diff
@@ -26,28 +26,3 @@ func.func public @streamer_add(%arg0 : memref<16xi64>, %arg1 : memref<16xi64>, %
 // CHECK-NEXT:    dart.yield %6 : !dart.stream<i64>
 // CHECK-NEXT:  }) : (index, index, index) -> ()
 
-// -----
```
jorendumoulin, line 29: turns out this was an illegal testcase, and I only found out in this test. this code never would have worked
this test is testing too much at once, it should be split up in the separate dart components

snaxc/transforms/convert_dart_to_snax_stream.py
```diff
@@ -36,29 +38,82 @@ def match_and_rewrite(self, op: dart.AccessPatternOp, rewriter: PatternRewriter)
...
+                stride, bound = next(access_iter)
+            else:
+                stride, bound = TCDM_BANK_WIDTH, (stride * bound) // TCDM_BANK_WIDTH
+
             # fill up all spatial strides
-            for _ in [x for x in template[0].bounds if x is not None]:
-                # FIXME: provide more general solution for new spatial streamer_config
-                # configuration, this works in all current cases and layouts but is far from generally correct.
-                spatial_strides = [8]
-                stride, bound = next(access_iter, (None, None))
```
jorendumoulin, line 61: this is the FIXME that is solved in this PR

snaxc/transforms/convert_dart_to_snax_stream.py
```diff
@@ -36,29 +38,82 @@ def match_and_rewrite(self, op: dart.AccessPatternOp, rewriter: PatternRewriter)
...
             upper_bounds: list[int] = []
 
+            # Fetch the first stride
+            stride, bound = next(access_iter)
+
+            # TCDM takes 8 contiguous bytes minimum
+            if stride * bound == TCDM_BANK_WIDTH:
+                stride, bound = next(access_iter)
+            else:
+                stride, bound = TCDM_BANK_WIDTH, (stride * bound) // TCDM_BANK_WIDTH
```
JosseVanDelm, line 90: This execution path takes one less value from the iterator. Is this expected?

snaxc/transforms/convert_dart_to_snax_stream.py
```diff
@@ -36,29 +38,82 @@ def match_and_rewrite(self, op: dart.AccessPatternOp, rewriter: PatternRewriter)
...
+                    # for contiguous access! Because the streamer is larger than the stride, it will
+                    # access more elements of the next stride, this is the applied stride. This should
+                    # correspond with the next stride for correct operation.
+                    assert spat_size % bound == 0
+                    applied_stride = stride * bound
+                    applied_bound = spat_size // bound
+                    next_stride, next_bound = next(access_iter)
+                    if applied_stride != next_stride:
+                        raise RuntimeError(
+                            "Non-contiguous access is not possible for this streamer configuration"
```
JosseVanDelm, line 109: Maybe something like ... "tried to apply a stride of {applied_stride}, but this is not available within ... which requires a minimum stride of {next_stride}" ? Would that make sense?

snaxc/transforms/convert_dart_to_snax_stream.py
```diff
@@ -36,29 +38,82 @@ def match_and_rewrite(self, op: dart.AccessPatternOp, rewriter: PatternRewriter)
...
+            for spat_size in streamers[operand].spatial_dims:
+                spatial_strides.append(stride)
+                if bound == spat_size:
+                    # nice, strides correspond with streamer dimension
+                    stride, bound = next(access_iter)
+                elif bound < spat_size:
+                    # caution! strides of pattern don't nicely overlap with streamers, only possible
+                    # for contiguous access! Because the streamer is larger than the stride, it will
+                    # access more elements of the next stride, this is the applied stride. This should
+                    # correspond with the next stride for correct operation.
```
JosseVanDelm, line 102: This comment is not clear for me

JosseVanDelm (commented):
Sorry I but I find the current code quite difficult to read, can we try to modularize it a bit more? maybe put more things in functions, add more big comments on what is happening?

JosseVanDelm (approved):

snaxc/transforms/convert_dart_to_snax_stream.py
jorendumoulin: yes

PR #390, opened 2025-04-24
transform: (dart-scheduler) enable schedule idx in scheduler

jorendumoulin:
for now, the scheduler just returns the first valid schedule.
while the selection of schedule should be a bit smarter, this allows for a selection of a specific index in the list of valid indeces as shortcut or to explore multiple of them

### DIFF
--- a/snaxc/ir/dart/scheduler.py
+++ b/snaxc/ir/dart/scheduler.py
@@ -129,7 +129,11 @@ def scheduler(
         # defaulting to pure output stationary schedules for now
         is_pure_output_stationary
     ],
+    schedule_idx: int | None = None,
 ) -> Schedule:
     # for now just return the first result of the backtracking
+    if schedule_idx is not None:
+        all = list(scheduler_backtrack(template, schedule, extra_checks=extra_checks))
+        return all[schedule_idx]
     result = next(scheduler_backtrack(template, schedule, extra_checks=extra_checks))
     return result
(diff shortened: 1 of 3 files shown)


snaxc/ir/dart/scheduler.py
```diff
@@ -129,7 +129,11 @@ def scheduler(
         # defaulting to pure output stationary schedules for now
         is_pure_output_stationary
     ],
+    schedule_idx: int | None = None,
 ) -> Schedule:
     # for now just return the first result of the backtracking
+    if schedule_idx is not None:
+        all = list(scheduler_backtrack(template, schedule, extra_checks=extra_checks))
```
JosseVanDelm, line 136: Probably overkill, though:

Seems a bit excessive to get all schedules and then just return one.
Maybe instead:
```python
if schedule_idx is not None:
    for i, schedule in enumerate(scheduler_backtrack(template, schedule, extra_checks=extra_checks)):
        if i == schedule_idx:
            return schedule
```
If this is called in a function multiple times, it actually might make sense to cache the outcome.
```python
from functools import lru_cache

@lru_cache(maxsize=None)
def get_schedule_at_idx(template, schedule, schedule_idx, extra_checks=None):
    if schedule_idx is not None:
        for i, sched in enumerate(scheduler_backtrack(template, schedule, extra_checks=extra_checks)):
            if i == schedule_idx:
                return sched
    return None
```

JosseVanDelm (approved):
Cool!

snaxc/ir/dart/scheduler.py
jorendumoulin: this is a bit overkill imo, during debugging it is actually very useful to get access to all of them to investigate
this is more of a testing feature anyway, so nothing performance critical

PR #391, opened 2025-04-24
transform: (set-memory-layout) only tile layout if all patterns remain affine

jorendumoulin:
when choosing an automatic layout, the selection to tile or not to tile is a bit more complex.
As the streamers can only have affine access patterns, the resulting patterns after tiling must remain affine, which is what this one checks.

### DIFF
--- a/snaxc/transforms/set_memory_layout.py
+++ b/snaxc/transforms/set_memory_layout.py
@@ -99,22 +99,25 @@ class AddCyclicMemoryLayout(RewritePattern):
                 )
 
                 # can we further tile the layout according to the remaining size?
-                # only apply tiling if the entire size is nicely divisible by the tile size for now
                 to_tile = self.tiled_layout
 
-                # only apply tiling if entire size is nicely divisible by the tile size for now (not strictly necessary)
-                if size_remaining % schedule_bound != 0:
-                    to_tile = False
-
-                # only apply tiling if all access patterns remain hyperrectangular
-                # for example if there is one dimensions that accesses with stride=1, bound=3 and another dim with
-                # stride=1, bound=8
-                # we cannot tile for either 8 or 3 because then the other pattern is no longer affine
-                for stride, bound in zip(
-                    schedule.pattern.A[accessed_dim, :], schedule.bounds
-                ):
-                    if stride % schedule_bound != 0 and bound != schedule_bound:
+                if to_tile:
+                    # only apply tiling if entire size is nicely divisible by the tile size for now
+                    # (this isn't strictly necessary but might break some things)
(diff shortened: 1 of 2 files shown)


tests/filecheck/transforms/set-memory-layout.mlir
```diff
@@ -23,3 +23,29 @@ func.func @gemm(%arg0 : memref<16x16xi8, "L1">, %arg1 : memref<16x16xi8, "L1">,
...
+      ^1(%arg3 : i8, %arg4 : i8, %arg5 : i32):
+        %8 = kernel.mac %arg3, %arg4 : i8, i8 -> i32
+        dart.yield %8 : i32
+      }) : (!dart.stream<i8>, !dart.stream<i8>) -> !dart.stream<i32>
+      dart.yield %7 : !dart.stream<i32>
+    }) : (memref<1x16x18x18xi8, "L1">, memref<16x16x3x3xi8, "L1">, memref<1x16x16x16xi32, "L1">) -> ()
+  func.return
+}
+
+// TILED:      %3 = "snax.layout_cast"(%0) : (memref<1x16x18x18xi8, "L1">) -> memref<1x16x18x18xi8, #tsl.tsl<[1] -> (5184), [2, 8] -> (2592, 1), [18] -> (144), [18] -> (8)>, "L1">
```
jorendumoulin, line 45: without the extra check in this PR the H dimension would be tiled as [6, 3], which is impossible to then realize with the streamers

snaxc/transforms/set_memory_layout.py
```diff
@@ -100,7 +100,22 @@ def match_and_rewrite(self, op: dart.ScheduleOp, rewriter: PatternRewriter):
...
+
+                # only apply tiling if all access patterns remain hyperrectangular
+                # for example if there is one dimensions that accesses with stride=1, bound=3 and another dim with
+                # stride=1, bound=8
+                # we cannot tile for either 8 or 3 because then the other pattern is no longer affine
+                for stride, bound in zip(
+                    schedule.pattern.A[accessed_dim, :], schedule.bounds
+                ):
+                    if stride % schedule_bound != 0 and bound != schedule_bound:
+                        to_tile = False
```
JosseVanDelm, line 117: This check seems redundant if you already decided you don't want to tile, right?
```suggestion
                else:
                    for stride, bound in zip(
                        schedule.pattern.A[accessed_dim, :], schedule.bounds
                     ):
                        if stride % schedule_bound != 0 and bound != schedule_bound:
                             to_tile = False
```

JosseVanDelm (approved):
Nice! Please see my remark

PR #392, opened 2025-04-24
transforms: (set-memory-layout) fill in empty strides

jorendumoulin:
if there are empty strides, just fill in an empty one to avoid weird cases


jorendumoulin:
part of #393

PR #393, opened 2025-04-24
add dart schedule canonicalization

jorendumoulin:
canonicalize operation schedules before applying the scheduler to avoid duplicate results

For example a for loop with bound 1 is useless, but rotating this one around results in several valid schedules that are actually equal. Therefore there is a canonicalization step before applying the scheduler

### DIFF
--- a/snaxc/transforms/dart/dart_scheduler.py
+++ b/snaxc/transforms/dart/dart_scheduler.py
@@ -49,6 +49,9 @@ class AutoflowScheduler(RewritePattern):
         )
         schedule = scheduler(template, schedule, schedule_idx=self.schedule_idx)
 
+        schedule = schedule.canonicalize()
+        schedule = scheduler(template, schedule)
+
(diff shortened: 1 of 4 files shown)


snaxc/transforms/dart/dart_scheduler.py
```diff
@@ -46,6 +46,8 @@ def match_and_rewrite(self, op: dart.OperationOp, rewriter: PatternRewriter):
             SchedulePattern(schedule_bounds, pattern.data)
             for pattern in op.patterns.data
         )
+
+        schedule = schedule.canonicalize()
```
JosseVanDelm, line 50: Is there a reason to not always canonicalize?
Can we not put this inside the dart scheduler instead of calling it from the outside?

jorendumoulin: i want to be able to test the scheduler without it canonicalizing away possible things

PR #394, opened 2025-04-24
transforms: (snax-copy-to-dma) add ignore transform option

jorendumoulin:
sometimes you just want to ignore all transformations made by the dma because it makes everything very slow, this gives you the option to replace it by a regular 1D transfer.
The data will be wrong, but it can be helpful during debugging

### DIFF
--- a/snaxc/transforms/snax_copy_to_dma.py
+++ b/snaxc/transforms/snax_copy_to_dma.py
@@ -1,3 +1,5 @@
+from dataclasses import dataclass
+
 from xdsl.context import Context
 from xdsl.dialects import arith, builtin, func, scf
 from xdsl.dialects.arith import AddiOp, ConstantOp, MuliOp
@@ -160,8 +162,11 @@ def extract_offset(memreftype: MemRefType):
     return 0
 
 
+@dataclass
 class TransformDMA(RewritePattern):
     """Look for memref copy operations with TSL layout and insert snitch DMA calls"""
+
+    test_ignore_transform: bool = False
 
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: CopyOp, rewriter: PatternRewriter):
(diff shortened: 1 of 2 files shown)


snaxc/transforms/snax_copy_to_dma.py
```diff
@@ -485,16 +490,23 @@ def match_and_rewrite(self, op: CopyOp, rewriter: PatternRewriter):
...
 
+    test_ignore_transform: bool | None = False
+
     def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
         PatternRewriteWalker(MatchSimpleCopy()).rewrite_module(op)
-        PatternRewriteWalker(TransformDMA()).rewrite_module(op)
+
+        ignore = self.test_ignore_transform
+        if ignore is None:
+            ignore = False
```
JosseVanDelm, line 508: Is this not redundant when the default is specified as = False on line 501?

tests/filecheck/transforms/copy-to-dma-ignore-transform.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: snax-opt --split-input-file %s -p snax-copy-to-dma{test-ignore-transform=true} | filecheck %s
+
+// a very complext transformation:
```
JosseVanDelm, line 3:
```suggestion
// a very complex transformation:
```

snaxc/transforms/snax_copy_to_dma.py
```diff
@@ -485,16 +490,23 @@ def match_and_rewrite(self, op: CopyOp, rewriter: PatternRewriter):
         rewriter.replace_op(op, outermost_for_loop)
 
 
+@dataclass(frozen=True)
 class SNAXCopyToDMA(ModulePass):
     """
     This pass translates memref copies to snitch DMA calls.
     """
```
JosseVanDelm, line 497: The motivation for this PR seems useful to be part of the documentation of this pass. How about:
```suggestion
    """
    This pass translates memref copies to snitch DMA calls.
    
    If you want to ignore all transformations made by the dma because it makes everything very slow, the    test_ignore_transform gives you the option to replace it by a regular 1D transfer. 
    Note that this renders the data incorrect, yet it can be helpful during debugging
    """
```

JosseVanDelm (approved):
Seems useful! Thanks!

snaxc/transforms/snax_copy_to_dma.py
jorendumoulin: unfortunately not, not setting the variable will initialize it to none

PR #395, opened 2025-04-24
dart scheduler: add memory flexibility test 

jorendumoulin:
this PR adds a check to the scheduler wheter a given schedule can be achieved, given the limited flexibility of the tcdm system. there will always have to be 8 int8 values packed together, and this check checks wether the schedule will access them at a finer granularity than this 8.

### DIFF
--- a/snaxc/ir/dart/scheduler.py
+++ b/snaxc/ir/dart/scheduler.py
@@ -122,12 +122,34 @@ def is_pure_output_stationary(template: Template, schedule: Schedule):
     return first_reduction_idx > last_parallel_idx
 
 
+def is_memory_flexible_enough(template: Template, schedule: Schedule):
+    """
+    Checks whether the TCDM flexibility is sufficient to actually execute
+    the schedule.
+
+    There must be one spatial stride of 1 that doesn't need more fine-grained
+    temporal access within one bank, such that that dimension can be packed together.
+    """
+    # this check only makes sense if there are temporal dimensions:
+    if not schedule.num_dims > template.num_dims:
+        return True
+    for s in schedule:
+        # is there temporary fine-grained access for this dimension?
+        temporal = (s.pattern.A[:, 0 : -template.num_dims] % 8).any(axis=1)
+        # is the dimension spatially unrolled?
+        spatial = (s.pattern.A[:, -template.num_dims :] == 1).any(axis=1)
+        if (False, True) not in zip(temporal, spatial):
+            return False
+    return True
+
+
 def scheduler(
     template: Template,
     schedule: Schedule,
     extra_checks: Sequence[Callable[[Template, Schedule], bool]] = [
         # defaulting to pure output stationary schedules for now
-        is_pure_output_stationary
+        is_pure_output_stationary,
+        is_memory_flexible_enough,
     ],
 ) -> Schedule:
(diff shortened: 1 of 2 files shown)


snaxc/ir/dart/scheduler.py
```diff
@@ -122,12 +123,38 @@ def is_pure_output_stationary(template: Template, schedule: Schedule):
...
+):
+    """
+    Checks whether the TCDM flexibility is sufficient to actually execute
+    the schedule.
+
+    There must be one spatial stride of 1 that doesn't need more fine-grained
+    temporal access within one bank, such that that dimension can be packed together.
+    """
+    TCDM_BANK_WIDTH = 8
+    # this check only makes sense if there are temporal dimensions:
```
JosseVanDelm, line 137: I'm a bit confused by this phrasing, how about this?
```suggestion
    # If there are no temporal dimensions, this check is not necessary and the memory is for sure flexible enough:
```

snaxc/ir/dart/scheduler.py
```diff
@@ -122,12 +123,38 @@ def is_pure_output_stationary(template: Template, schedule: Schedule):
...
+    template: Template, schedule: Schedule, element_sizes: Sequence[int]
+):
+    """
+    Checks whether the TCDM flexibility is sufficient to actually execute
+    the schedule.
+
+    There must be one spatial stride of 1 that doesn't need more fine-grained
+    temporal access within one bank, such that that dimension can be packed together.
+    """
+    TCDM_BANK_WIDTH = 8
```
JosseVanDelm, line 136: Should we make this not hardcoded, with a default?

snaxc/ir/dart/scheduler.py
```diff
@@ -122,12 +123,38 @@ def is_pure_output_stationary(template: Template, schedule: Schedule):
...
 
+def is_memory_flexible_enough(
+    template: Template, schedule: Schedule, element_sizes: Sequence[int]
+):
+    """
+    Checks whether the TCDM flexibility is sufficient to actually execute
+    the schedule.
+
+    There must be one spatial stride of 1 that doesn't need more fine-grained
+    temporal access within one bank, such that that dimension can be packed together.
```
JosseVanDelm, line 134: Possibly very dumb question: Where does this packing together happen?
Should you mention what makes sure this is okay in the end?

JosseVanDelm (commented):
Do my comments make sense here?
Also, it seems useful for the scheduler to have access to the AccContext to insert these things automatically (yet maybe future work?))

snaxc/ir/dart/scheduler.py
jorendumoulin: The packing is implicit and a result of the selected datalayout, in our case mostly the channels are packed together in groups of 8 most of the time.

snaxc/ir/dart/scheduler.py
jorendumoulin: There is no infrastructure to put this value yet, so this is the reason I just made it a variable here.
I agree this should live somewhere better, AccContext or something, but I felt like this is a separate (future) issue

jorendumoulin: There are also no systems yet with another value than 8

PR #401, opened 2025-05-06
Adding an MLP program

rgantonio:
In this PR, let's try out an MLP where you can vary the input size, hidden layers, and output sizes. 

Things you will find:
⭐  A *golden* function to produce MLP outputs
⭐  The MLIR programming function
⭐ The other things required to run on a certain SNAX cluster architecture

🌋 : I purposely separated this so I can modularize and make comparisons. I know in other examples we can combine them into one, but I did this for my *learning* purposes.

⚠️ Right now this is not "perfect" as it complains about the sign extension due to the `i32` output of the matmul operation. We need a scaling capability.

For now, I assumed that the outputs are indeed 32 bit but setting the types to 8bits complains more. So I get this error after running `snax-opt`:

```bash
 tensor<1x3xi32>
    %21 = "tensor.empty"() : () -> tensor<1x3xi32>
    %22 = "linalg.quantized_matmul"(%9, %2, %7, %7, %21) <{operandSegmentSizes = array<i32: 4, 1>}> ({
    ^1(%23 : i32, %24 : i8, %25 : i32, %26 : i32, %27 : i32):
      %28 = "arith.extsi"(%23) : (i32) -> i32
      ^^^^^^^^^^^^^^^^^^^-----------------------------------------------------------------------
      | Operation does not verify: Destination bit-width must be larger than the input bit-width
      ------------------------------------------------------------------------------------------
      %29 = "arith.subi"(%28, %25) <{overflowFlags = #arith.overflow<none>}> : (i32, i32) -> i32
      %30 = "arith.extsi"(%24) : (i8) -> i32
      %31 = "arith.subi"(%30, %26) <{overflowFlags = #arith.overflow<none>}> : (i32, i32) -> i32
      %32 = "arith.muli"(%29, %31) <{overflowFlags = #arith.overflow<none>}> : (i32, i32) -> i32
      %33 = "arith.addi"(%27, %32) <{overflowFlags = #arith.overflow<none>}> : (i32, i32) -> i32
```

Something to try out is the `xdsl.dialects.quant`. I found it here: https://mlir.llvm.org/docs/Dialects/QuantDialect/
It looks like (from what I understood) it can do the requantization steps. 

🌌**Solution**
Solution was to simply use the matMulOp rather than the quantizedMatMulOp. This way we can get away without the need for scaling yet. Moreoever, we use the shift scaling available so far.

⚠️ ~~I know this isn't formatted yet haha~~
⚠️ I hope to learn from what you guys think 💯

### DIFF
--- a/.github/workflows/build-run-kernel-snake.yml
+++ b/.github/workflows/build-run-kernel-snake.yml
@@ -20,4 +20,4 @@ jobs:
         working-directory: kernels/${{ matrix.kernel }}
     strategy:
       matrix:
-        kernel: [alloc, conv, simple_copy, transform_copy, gemm, rescale, gemmini, streamer_alu, streamer_matmul, tiled_add]
+        kernel: [alloc, conv, simple_copy, transform_copy, gemm, rescale, gemmini, streamer_alu, streamer_matmul, tiled_add, mlp]
--- a/kernels/mlp/Snakefile
+++ b/kernels/mlp/Snakefile
@@ -1,0 +1,67 @@
+from util.snake.configs import get_snax_gemmx_config
+
+config = get_snax_gemmx_config()
+config["snaxoptflags"] = ",".join(
+    [
+        "preprocess",
+        "convert-linalg-to-kernel",
+        "insert-accfg-op{accelerator=snax_gemmx}",
+        "dispatch-kernels",
+        "convert-linalg-to-dart",
+        "dart-fuse-operations",
+        "snax-bufferize",
+        "alloc-to-global",
+        "set-memory-space",
+        "dart-scheduler",
+        "set-memory-layout",
+        "realize-memref-casts",
+        "insert-sync-barrier",
+        "dispatch-regions{nb_cores=2}",
+        "dart-layout-resolution",
+        "convert-dart-to-snax-stream",
+        "convert-linalg-to-accfg",
+        "convert-accfg-to-csr",
+        "snax-copy-to-dma",
+        "memref-to-snax",
+        "snax-to-func",
+        "clear-memory-space",
+        "postprocess",
+    ]
+)
+
+
+
+module snax_rules:
+    snakefile:
+        "../../util/snake/snax.smk"
+    config:
+        config
+
+
+use rule * from snax_rules as snax_*
+
+
+files = ["mlp"]
+
+
+# Rules
+rule all:
+    input:
+        expand("{file}_traces.json", file=files),
+
+
+rule generate_mlp:
+    output:
+        "mlp.mlir",
+    script:
+        "mlp.py"
+
+
+rule link_snax_binary:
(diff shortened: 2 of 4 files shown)


jorendumoulin:
Nice! 

It's not recommended to use the quant dialect, last I heard, it is kind of deprecated

Turns out it's quite tricky to have a quantized matmul op with i8 outputs, so i replaced it to just use a matmul op, no need to have zero constants then!

Now it generates valid code for 4 cascaded matmuls, I'll get to work having this compile succesffully!

Thank you, this is going to be great 🥳

rgantonio:
@jorendumoulin nice! cool! Thanks a lot too 😄 

I thought the `quantizedMatMulOp` was our go to haha and I got worried that the `matMulOp` had to have extra arguments for defining the affine address if we are to transpose the matrices. I was about to try the `matMulOp` too haha

Should we worry about the part that says "Performs a matrix multiplication of two 2D inputs without broadcast or transpose."?

https://mlir.llvm.org/docs/Dialects/Linalg/#linalgmatmul-linalgmatmulop

jorendumoulin:
@rgantonio the testcase is working!

I update your python file such that the scaling corresponds to what is now used by default for an i8 output without explicit rescaling (`>>9`), and the results are correct!

I have also added the kernel to the CI already

rgantonio:
@jorendumoulin Yay thank you! I will tinker on it a bit on my end too just so I understand what happened 😄 

I'm starting to think we can develop toy models for these. Like expanding later to other "deeper" networks. The more examples we have, the more we can generalize!

rgantonio:
Wait! haha I realized that we do need the transpose stuff. Because when I change the hidden layers to something like [8,16,16,8] it now complains that at the part where the shape changes, it does not fit the multiplication.

jorendumoulin:
I found an issue when everything is set to 16, should be resolved now!

For the [8, 16, 8] case I believe not all dimensions are set correctly yet in `mlp.py`, which cause mlir to complain.
I'll see if I can make the error reporting better for this case

rgantonio:
@jorendumoulin thanks! I will try to check about the [8,16,8] stuff too probably over the weekend. Maybe it was my mistake to too save the sizes incorrectly.

rgantonio:
@jorendumoulin I know the problem! I have a fix 😄 wait for it hehe

rgantonio:
@jorendumoulin it's beautiful!!

Can now do different input/output sizes and the current setting of hidden layers is like [8,16,32,16,8]! It worksssss!!

jorendumoulin:
hooray!

jorendumoulin (approved):

rgantonio:
@jorendumoulin do I have permission to just squash and merge since it's approved now? haha

jorendumoulin:
yes!

PR #549, opened 2025-07-30
accelerators: (rescale_down_extension) Add RescaleDownExtension and kernel

JonasCrols:
adds functionality for the RescaleDownExtension and a kernel that tests it. Also makes minor changes to the StreamerExtension class to bring more flexibility to the subclasses


snaxc/accelerators/streamers/extensions/rescale_down_extension.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        # returns the number of input tensors in the rescale down op, currently limited to 2
+        assert isinstance(op, kernel.RescaleOp), "Operation must be a RescaleOp"
+
+        return [
+            op.input_zp.value.data,
+            int.from_bytes(op.multiplier.data.data[0:4], byteorder="little", signed=False),
+            op.output_zp.value.data,
+            int.from_bytes(op.shift.data.data[0:4], byteorder="little", signed=False),
+        ]
```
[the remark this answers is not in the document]
jorendumoulin: of `next(op.multiplier.iter_values())`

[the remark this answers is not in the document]
jorendumoulin: zoiets zou moeten bestaan denk ik, die int conversie hoort hier niet

snaxc/accelerators/streamers/extensions/maxpool_extension.py
```diff
@@ -28,3 +32,21 @@ def get_csr_values(self, op: kernel.KernelOp) -> Sequence[int]:
...
         return [1]  # TODO: Replace with actual CSR values for MaxPool
+
+    def get_template(self, op: kernel.KernelOp) -> Template:
+        """
+        Returns the template for this DMA extension.
+        """
+        template = [AffineMap.from_callable(lambda y: (y,))] * 2  # TODO: check if this is correct
+        template_bounds = (16,)
+        return Template(TemplatePattern(template_bounds, tp) for tp in template)
+
```
jorendumoulin, line 43: this template should be the default for a streamer extension, only in rare cases will this be different

snaxc/accelerators/snax_xdma.py
```diff
@@ -285,13 +298,40 @@ def get_xdma_streamer_launch_dict(self, base_addr: int = 0x3C0) -> tuple[int, di
...
+                            ):
+                                return ext.get_template(kernel_op)
+        return Template(
+            [
+                TemplatePattern(
+                    (16,),
+                    AffineMap.from_callable(lambda y: (y,)),  # Default template
+                )
+            ]
+        )
```
jorendumoulin, line 320: instead of doing the nested if approach, use asserts. There shouldn't be a template returned by default, only the correct ops may yield a valid template

snaxc/accelerators/snax_xdma.py
```diff
@@ -285,13 +298,40 @@ def get_xdma_streamer_launch_dict(self, base_addr: int = 0x3C0) -> tuple[int, di
...
+            kernel_op = kernel_op.body.block.first_op
+            if isinstance(kernel_op, KernelOp):
+                for streamer in self.streamer_config.data.streamers:
+                    for ext in streamer.opts:
+                        if isinstance(ext, StreamerExtension):
+                            if ext.supported_kernel is not None and isinstance(
+                                kernel_op, ext.supported_kernel.kernel_type
+                            ):
+                                return ext.get_streamers(streamer_config=self.streamer_config.data)
+        # If no specific extension is found, return the default streamers
```
jorendumoulin, line 333: same here for the nested ifs

snaxc/accelerators/snax_xdma.py
jorendumoulin: if we yield something default, something else will break down the line and it will be much more difficult to debug then, so it should fail at this point

PR #551, opened 2025-08-05
accelerators: (rescale_up_extension) Add rescale up extension

JonasCrols:
This PR:
* adds a new xDMA extension: rescale_up
* adapts the kernelOp.Rescale to allow 32-bit control values

PR #552, opened 2025-08-05
transforms: (combine-rescale) Add CombineRescale Pass

JonasCrols:
This PR adds functionality and tests for the CombineRescale pass which takes 2 subsequent rescales and transforms them into a single operation.


JonasCrols:
Temporarily closed as a small change from PR #551 is necessary

PR #553, opened 2025-08-07
accelerators: (maxpool) Complete MaxPool Extension

JonasCrols:


JonasCrols:
As xDSL needs to be updated for the maxpool.py file to work, we temporily use a pre-made mlir-source file for the maxpool.mlir file. When xDSL is updated, the python file can be used again by uncommenting the relevant section in the snakefile of the poolinf-directory

kernels/pooling/maxpool.py
```diff
@@ -0,0 +1,127 @@
...
+        # print(golden_vals[:,:, 1])
+        # print(golden_vals.flatten().tolist())
+
+        # # Declare result tensor type
+        # empty_tensor = EmptyOp([], output_type)
+
+        # Specify the operation
+        result = MaxPool2DOp(
+            operands=[a.result],
+            result_types=[output_type],
```
jorendumoulin, line 68: just clean the comments up a bit

PR #557, opened 2025-08-13
kernels: (pooling) Add AvgPool operation

JonasCrols:
This PR adds functionality for the AvgPool function from TOSA. AvgPooling is performed on the XDMA by chaining AddToLong Extension with Rescale Down


JonasCrols:
before merging this PR: the new version of xDSL needs to be available, the new hardware needs to be available.

PR #561, opened 2025-10-01
bump xdsl

jorendumoulin:


JosseVanDelm:
Hi,

It seems that we are running into issues because TOSA is moving to v1.0 of their spec introducing many breaking changes.
One of the breaking changes is that the tosa.fully_connected operation is completely gone.

At the same time Tensorflow is dropping all tflite APIs because they are moving all tflite stuff into litert.
Tensorflow is not updating their tflite to MLIR emission and hence now tensorflow is emitting stuff that is incompatible with upstream MLIR (the version xdsl uses already has moved on to v1.0 support for TOSA).

I'm including the IREE issue here because that seems like the best place to follow what is going on:
https://github.com/iree-org/iree/issues/19777

Some things I tried:
* Use tf 2.20 -> still emits old MLIR
* tf nightly --> API to do conversion is gone
* Use ARM's own converter tool at https://gitlab.arm.com/tosa/tosa-converter-for-tflite (this includes changes that are bleeding edge MLIR - i.e. it is too new for the version of MLIR we currently support).

IREE is advising people to not use tflite at the moment until things have settled on the upstream sides.
I'm not sure what this means and it is not clear what support can be expected in the future.

For this issue let's disable the test that is currently failing.
We can work on a monkey patch solution later. A better option is probably to rely on MLIR-to-MLIR tests instead of having them converted from a framework.

PR #564, opened 2025-11-03
Add custom syntax to phs dialect

JosseVanDelm:
Adds custom syntax to phs dialect to be easier on the eyes:
```mlir
builtin.module {
  phs.pe @myfirstaccelerator with %0, %1, %2, %3, %4, %5 (%6 : f32, %7 : f32) {
    %8 = phs.choose @_0 with %0 (%6 : f32, %7 : f32) -> f32 {
      0) arith.mulf
      1) arith.addf
    }
    %9 = phs.choose @_1 with %1 (%6 : f32, %8 : f32) -> f32 {
      0) arith.mulf
      1) arith.addf
    }
    %10 = phs.mux(%8 : f32, %9 : f32) -> f32 with %2
    %11 = phs.mux(%9 : f32, %7 : f32) -> f32 with %5
    %12 = phs.choose @_2 with %3 (%8 : f32, %11 : f32) -> f32 {
      0) arith.mulf
      1) arith.divf
    }
    %13 = phs.mux(%10 : f32, %12 : f32) -> f32 with %4
    phs.yield %13 : f32
  }
}

PR #566, opened 2025-11-04
Add support for arbitrary types and operation arity in choose_op

JosseVanDelm:
Adapts choose_ops to allow any type. Previously this was confined to certain float operations.
Also slightly changes mux syntax to be more in line with what has been developed in #564


JosseVanDelm:
:heart_decoration:

PR #577, opened 2025-12-01
add option for snaxc config for perfetto traces

jorendumoulin:


util/tracing/trace_to_perfetto.py
```diff
@@ -28,6 +31,13 @@ def worker(file: str, accelerator: str):
         accelerator_op = get_all_accelerators()[accelerator]().generate_acc_op()
         generators.append(SNAXAcceleratorEventGenerator(accelerator_op))
 
+    if accelerator_config is not None:
+        with open(accelerator_config) as f:
+            config = yaml.safe_load(f)
+        context = parse_config(config)
+        for registered_accelerator in context._registered_accelerators.values():
+            generators.append(SNAXAcceleratorEventGenerator(registered_accelerator().generate_acc_op()))
```
JosseVanDelm: Can you maybe add an accessor to context.registered_accelerators instead of calling the private variable? Does pyright allow this?.
I'm not sure why we are not just using the get_all_accelerators thing here, is it because the config parser can create new accelerators on the fly?

PR #598, opened 2026-02-10
Add initial support for floating point in PHS

JosseVanDelm:
Add support by calling into hardfloat library


JosseVanDelm:
This already works on main now

